Issue #6, opened 2022-06-23
Vue Component : Circuit Element Panel - GSoC'22


devartstar:
Changes to be made:-

- [x] Convert jQuery UI accordion to Vuetify component and fix styles.
- [x] Remove repeated DOM manipulation with strings for displaying panel icons.
- [x] Move the logic for searching of circuit elements to inside the panel-component.
- [ ] Refactor CSS for the Elements Panel.

PR #7, opened 2022-06-23
feat: circuit element panel component

devartstar:
Fixes: #6 
- Created PanelHeader Vue file - common for all panels which take in the title as a prop and display it.

- Converted the jQuery UI accordion -> [Vuetify expansion panel (variant: accordion)](https://next.vuetifyjs.com/en/components/expansion-panels/)

- Fixed some styles like to match the styles.

- Removed DOM manupulation as string for displaying icons in `setupElementLists()` function in `setup.js`.

- Fetching the data using lifecycle hook `onBeforeMount` and storing it in proper form to display in panel body using for loop.

- Moved `createElement()` function (which creats elemets to drag on the canvas) from `ux.js` to inside vue-componet

- Move the logic of searching elements inside vue-component

**To do:-** 
- [ ] Refactoring of CSS for circuit elements panel


src/components/Panels/PanelHeader.vue
```diff
@@ -0,0 +1,13 @@
...
+    <div class="panel-header">
+        {{ props.headerTitle }}
+        <span class="fas fa-minus-square minimize"></span>
+        <span class="fas fa-external-link-square-alt maximize"></span>
+    </div>
+</template>
+
+<script lang="ts" setup>
+const props = defineProps({
+    headerTitle: { type: String, default: 'Panel Header' },
```
tachyons, line 11: Can we make the texts translatable from the beginning itself, it will be difficult to translate later

devartstar: yes sure, I will try to do it.

devartstar: Worked on it, can you please review it.

https://user-images.githubusercontent.com/61665451/175772655-4800403c-62c9-4b73-a1e5-67e7bc319296.mp4

src/components/Extra.vue
```diff
@@ -978,8 +956,6 @@
     <!-- --------------------------------------------------------------------------------------------- -->
 </template>
 
-<script>
-export default {
-    name: 'Extra',
-}
+<script lang="ts" setup>
+import ElementsPanelVue from './Panels/ElementsPanel/ElementsPanel.vue'
```
abstrekt, line 960: lets not use the 'vue' suffix in component names?

src/components/Panels/PanelHeader.vue
```diff
@@ -0,0 +1,13 @@
+<template>
```
abstrekt, line 1: if this is a sharable component between the Panels, let's create a Panels/shared & move it to there, same flow for any shareable file.

src/components/Panels/ElementsPanel/ElementsPanel.vue
```diff
@@ -0,0 +1,186 @@
...
+            }
+        }
+    }
+    return finalResult
+}
+
+function createElement(elementName) {
+    if (simulationArea.lastSelected && simulationArea.lastSelected.newElement)
+        simulationArea.lastSelected.delete()
+    var obj = new modules[elementName]()
```
abstrekt, line 171: lets enable `https://eslint.org/docs/latest/rules/no-var`, wdyt?

src/components/Panels/ElementsPanel/ElementsPanel.vue
```diff
@@ -0,0 +1,186 @@
...
+function getImgUrl(elementName) {
+    const elementImg = new URL(
+        `../../../assets/img/${elementName}.svg`,
+        import.meta.url
+    ).href
+    return elementImg
+}
+
+var elementInput = ref('')
+function searchElements() {
```
abstrekt, line 138: Please make the element search case-insensitive. Currently, I can I'm not able to search for Hex using text "Hex", only "hex" is working.

src/components/Panels/ElementsPanel/ElementsPanel.vue
```diff
@@ -0,0 +1,186 @@
+<template>
+    <div
```
abstrekt, line 2: I can experience some lag when opening the panels when trying from the deploy preview, this doesn't seems to be the case on the production instance, can you check?

abstrekt:
On initial load, the minimize panel icon on the header seems broken -
image

There is some padding missing on the button for the last element
image

src/components/Panels/ElementsPanel/ElementsPanel.vue
```diff
@@ -0,0 +1,186 @@
...
+import simulationArea from '#/simulator/src/simulationArea'
+import { uxvar } from '#/simulator/src/ux'
+import modules from '#/simulator/src/modules'
+import { onBeforeMount, ref } from 'vue'
+var panelData = []
+window.elementHierarchy = metadata.elementHierarchy
+window.elementPanelList = []
+
+import { useI18n } from 'vue-i18n'
+const { t, locale } = useI18n()
```
abstrekt, line 112: lets make this function global? so that we don't need to import it everywhere?

abstrekt (changes requested):

src/styles/css/main.stylesheet.css
```diff
@@ -500,6 +501,44 @@ input[type='text']:focus {
     position: relative;
 }
 
+/* expansion panel styles - start */
+.v-expansion-panel {
```
abstrekt, line 505: we shouldn't overwrite library components styles globally, lets's move this to the components style tag, and wrap it inside a custom class?

devartstar:
> On initial load, the minimize panel icon on the header seems broken - 

fixed it, both the maximize and minimize icons were overlapping initially on load giving that weird icon.

> There is some padding missing on the button for the last element

fixed it, margin-bottom to the last panel was not showing.

src/components/Extra.vue
devartstar: removed it.

src/components/Panels/PanelHeader.vue
devartstar: moved PanelHeader into a folder named shared.

Will also have to make a separate sharable component for `<v-expansion-panel-text>` 
Since it's repeated twice in `ElementsPanel.vue`
And maybe needed once again in `LayoutPanel.vue`

src/components/Panels/ElementsPanel/ElementsPanel.vue
devartstar: If we add this rule we might have to disable it for the `simulators folder`, else it will add up to the linting errors since lots of places use var.
or insead i will use `let` and `const` from now on.

src/components/Panels/ElementsPanel/ElementsPanel.vue
devartstar: Fixed it.

src/styles/css/main.stylesheet.css
devartstar: I will fix it next while refactoring the CSS for this component.

src/components/Panels/ElementsPanel/ElementsPanel.vue
devartstar: Seemed like it was a common problem for many.
Solved it using this [reference](https://github.com/vuetifyjs/vuetify/issues/10140#issuecomment-605815875)

src/components/Panels/PanelHeader.vue
tachyons: fantastic

devartstar:
Will refactor the CSS after Layout Elements Panel is also done. (many classes are the same)

Issue #8, opened 2022-07-01
Vue Component : Properties Panel - GSoC'22


devartstar:
Replace the properties panel done by DOM manupulation using string to a Vue Component.

Tasks to be done :-
- [x] Seperate the Properties panel into its own component.
- [x] Move the logic from `showProperties()` function inside Vue component to display the panel accordingly.
- [ ] There are 3 types of Properties Panel Body
> - [ ] 1. Properties of elements in Layout Mode
> - [ ] 2. Properties Panel showing Project Properties
> - [ ] 3. Properties Panel showing Circiut Element Properties
- [ ] Link each of the options in the properties panel to their functionality
- [ ] Refactoring of CSS

PR #9, opened 2022-07-02
feat: properties panel component

devartstar:
Fixes: #8 
Tasks to be done :-

- [x]  Seperate the Properties panel into its own component.
- [x]  Move the logic from showProperties() function inside Vue component to display the panel accordingly.
- [x]  There are 3 types of Properties Panel Body
> - [x]  1. Properties Panel showing Project Properties 
> - [x]  2. Properties Panel showing Circiut Element Properties
> - [x]  3. Properties of elements in Layout Mode
- [x]  Link each of the options in the properties panel to their functionality
- [ ]  Refactoring of CSS


devartstar:
@abstrekt @ayan-biswas0412 
I have completed the panel body when circuit elements are selected.
Can you please check once if all the properties panel functionality wrt circuit elements are working properly?
Also, can you review the code so far

ayan-biswas0412 (commented):
* The Testbench is not draggable
image

* The **5f** and **0** both elements become same element when importing to the simulator panel
image

devartstar:
@ayan-biswas0412 
> The Testbench is not draggable
I experimentally applied the drag functiontionality just to the elements panel by coding the function myself. In the meet, it was decided that we might use the drag functionality logic from VueUse, if we don't find a proper library, hence haven't applied drag to any other component.

> The 5f and 0 both elements become same element when importing to the simulator panel
Can you please describe the bug? I compared it with the current simulator and it seems to work the same.

ayan-biswas0412:
> @ayan-biswas0412
> 
> > The Testbench is not draggable
> 
> I experimentally applied the drag functiontionality just to the elements panel by coding the function myself. In the meet, it was decided that we might use the drag functionality logic from VueUse, if we don't find a proper library, hence haven't applied drag to any other component.
> 
> > The 5f and 0 both elements become same element when importing to the simulator panel
> 
> Can you please describe the bug? I compared it with the current simulator and it seems to work the same.
> 
>  icons.mp4

ok no issues then.
I re-checked that and it is not a bug 🎉
Good work @devartstar 🚀

devartstar:
There are 2 things that I need to fix -

- The CircuitName in the panel doesn't change on switching circuits
- After canceling layout mode the panel doesn't reappear

src/components/Panels/PropertiesPanel/PropertiesPanel.vue
```diff
@@ -0,0 +1,80 @@
...
+    <div
+        id="moduleProperty"
+        class="moduleProperty noSelect effect1 properties-panel draggable-panel draggable-panel-css guide_2"
+    >
+        <PanelHeader :header-title="$t('simulator.panel_header.properties')" />
+        <div class="panel-body">
+            <div id="moduleProperty-inner">
+                <div id="moduleProperty-header">{{ panelBodyHeader }}</div>
+                <PanelType1 v-if="panelType == 1" />
+                <PanelType2 v-if="panelType == 2" />
```
abstrekt, line 11: If at a time, only one of these PanelType is to be rendered, then it is best practice to chain v-if with v-else-if.

src/components/Panels/PropertiesPanel/PropertiesPanel.vue
```diff
@@ -0,0 +1,80 @@
...
+import HelpButton from '../Shared/HelpButton.vue'
+import { onMounted } from 'vue'
+import {
+    checkPropertiesUpdate,
+    hideProperties,
+    prevPropertyObjGet,
+    prevPropertyObjSet,
+} from '#/simulator/src/ux'
+import { layoutModeGet } from '#/simulator/src/layoutMode'
+import PanelType1 from './PanelTypes/PanelType1.vue'
```
abstrekt, line 39: Lets add proper names to `PanelType1` , `PanelType2`, `PanelType3`, & use alias for any import.

src/components/Panels/PropertiesPanel/PropertiesPanel.vue
```diff
@@ -0,0 +1,80 @@
...
+        <PanelHeader :header-title="$t('simulator.panel_header.properties')" />
+        <div class="panel-body">
+            <div id="moduleProperty-inner">
+                <div id="moduleProperty-header">{{ panelBodyHeader }}</div>
+                <PanelType1 v-if="panelType == 1" />
+                <PanelType2 v-if="panelType == 2" />
+                <PanelType3
+                    v-if="panelType == 3"
+                    :key="toRaw(propertiesPanelObj)"
+                    :data="toRaw(propertiesPanelObj)"
```
abstrekt, line 15: why was toRaw necessary here?

src/components/Panels/PropertiesPanel/PanelTypes/PanelType3.vue
```diff
@@ -0,0 +1,122 @@
...
+</template>
+
+<script lang="ts" setup>
+import { escapeHtml } from '#/simulator/src/utils'
+import InputGroups from '../../Shared/InputGroups.vue'
+import DropdownSelect from '../../Shared/DropdownSelect.vue'
+const props = defineProps({
+    data: { type: Object, default: undefined },
+})
+const obj = props.data
```
abstrekt, line 120: 1/ lets use proper variable names here,
2/ storing props item in a variable like this removes the reactivity, are we intentionally removing reactivity?
3/ you can do something like `const { obj } = toRefs(props)` to store props in a variable and retain reactivity.
4/ also, the template can directly access all props, we don't need to extract it in the setup for the template to access props

src/components/Panels/Shared/HelpButton.vue
```diff
@@ -0,0 +1,22 @@
...
+		</button>
+	</p>
+</template>
+
+<script lang="ts" setup>
+const props = defineProps({
+    data: { type: Object, default: undefined },
+})
+const obj = props.data
+var helplink = obj && obj.helplink
```
abstrekt, line 18: just some tips: you can checkout JS Optional Chaining feature.

abstrekt (commented):
Functionality issue -
* Delete Circuit is not functional
* When exiting layout mode, properties panel is no longer accessible

src/components/Extra.vue
```diff
@@ -24,29 +24,7 @@
```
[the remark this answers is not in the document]
devartstar: Okay, I will.
But I don't think any code will be left in this at the end.
All of them will be moved to their respective components.

src/components/Panels/PropertiesPanel/PropertiesPanel.vue
```diff
@@ -0,0 +1,80 @@
...
+    prevPropertyObjSet(simulationArea.lastSelected)
+    propertiesPanelObj.value = simulationArea.lastSelected
+    // there are 3 types of panel body for Properties Panel
+    // depending upon which is last selected
+    // 1. Properties Panel in Layout mode
+    // 2. Properties Panel showing Project Properties
+    // 3. Properties Panel showing Circiut Element Properties
+
+    if (layoutModeGet()) {
+        panelType.value = 1
```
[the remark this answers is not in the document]
devartstar: This would be the Layout Panel

image

and the properties of the layout elements on selecting

image

src/components/Panels/PropertiesPanel/PropertiesPanel.vue
devartstar: Okay will change it

src/components/Panels/PropertiesPanel/PropertiesPanel.vue
devartstar: assigning an object to data property it converts into proxy object.
I found out Vue 3 is using [ES6 Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to make an objects reactive
I was not able to compare `propertiesPanelObj` with `simulationArea.lastselected` to prevent updating panel type every time.
So used `toRaw()` to return the original object from proxy maintaining reactivity.

src/components/Panels/Shared/HelpButton.vue
devartstar: Thanks, I implemented it for help link

src/components/Panels/PropertiesPanel/PropertiesPanel.vue
devartstar: Yes, I renamed it and also restructured the component folder structure.

src/components/Extra.vue
```diff
@@ -978,8 +946,7 @@
     <!-- --------------------------------------------------------------------------------------------- -->
 </template>
 
-<script>
-export default {
-    name: 'Extra',
-}
+<script lang="ts" setup>
+import ElementsPanel from './Panels/ElementsPanel/ElementsPanel.vue'
```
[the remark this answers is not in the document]
devartstar: yes, I replaced it in all the places inside the properties panel.

devartstar:
> Functionality issue -
> 
> * Delete Circuit is not functional
> * When exiting layout mode, properties panel is no longer accessible

Fixed them.

Issue #27, opened 2022-10-12
Drop JQuery dependency


tachyons:
Though we removed jquery ui, we still have jquery in our codebase. which need to be removed and use native DOM methods

disha1202:
Hi @tachyons, 
I would like to work on this.

tachyons:
@disha1202 You can start working on this, but since jquery is used in several files, PR will be very huge. So I suggest you create PRs with changes per file. Also multiple people can work on this if you cab mention which file/files you want to work on

disha1202:
Sure @tachyons, I would like to start with Verilog2CV.js file.

snehalshah8429:
@tachyons can I work on this?

snehalshah8429:
@tachyons  I would like to contribute on this

snehalshah8429:
@disha1202 do you need any help in this project?

PRO-BROS:
Very well. Facing the same issue but now resolved on my [site](https://prhocoding.com/)

disha1202:
@tachyons, I've removed JQuery dependency in Verilog2CV.js file.
Please take a look.

GarvitSinghal47:
@tachyons and @devartstar as there are multiple js files in which there is still jquery present in them I would like to work and help remove jquery from some of the files.

GarvitSinghal47:
Hey @tachyons  as i discussed in the slack i am currently working on circuit.js file.

niladrix719:
i am working on action.js

antriksh-9:
Hey @Arnabdaz @Prerna-0202 I have worked on PR #226.

antriksh-9:
I am working on testbench.js file.

antriksh-9:
I am working on tutorials.js file.

antriksh-9:
I am working on ux.js file.

antriksh-9:
I am working on canvasApi.js file.

tachyons:
Note: when you move transitions like fade from JQuery, use [Vue transitions](https://vuejs.org/guide/built-ins/transition) instead of trying to implement it vanilla JS.

PR #34, opened 2022-12-29
refactor: removed JQuery dependency from verilog.js file

disha1202:
Fixes: #27 

Removed JQuery dependency from verilog.js file


disha1202:
@devartstar Please take a look.

PR #53, opened 2023-01-07
draft PR: search functionality in custom shortcuts panel

### LINKED ISSUE #48: Add a search functionality to Custom keyboard shortcuts section
**Is your feature request related to a problem? Please describe.**

- It would be much easier to search for a certain shortcut if a user wants to alter or view one than to scroll all the way to the bottom.

**Describe the solution you'd like**

- Add a search area and a button to filter out the shortcut.

**Additional context**
**ScreenShot:**
Screenshot from 2023-01-05 07-23-04

**Are you working on this? (Yes/No)**
**YES**

palashb01:
Fixes #48 

**Changes:**

- a search bar in the custom shortcut panel.

**Demo Video:**
circuitverse_draft_pr

**Problem:**

- Currently to check whether the given custom input is  assigned to something  is a procedure where, inside `customShortcuts.vue` a function 
` warnOverride($('#pressedKeys').text(), targetPref.value)` is called where `targetPref` send something like this:

Screenshot from 2023-01-07 08-01-28

- after this inside `actions.js` file a function is called [ref](https://github.com/CircuitVerse/cv-frontend-vue/blob/043093ba49fd9224e6d195d4d4289bb4504cc7ec/src/simulator/src/hotkey_binder/model/actions.js#L110), which iterate through all the present children of div and match whether it has been assigned already or not.
- with the new changes if you search something then the number of shortcuts will reduce on the screen which reduces the child div;s because of this:

- **if you try to assign a shortcut that has already been assigned to something else( which is not shown on the screen) then it won't warn and just assign that shortcut to the given command.** 

> I was wondering if there could be a way to implement this feature of checking the assigned shortcuts differently instead of children div's then this search functionality could work much better.

cc: @devartstar @tachyons

### DIFF
--- a/src/components/DialogBox/CustomShortcut.vue
+++ b/src/components/DialogBox/CustomShortcut.vue
@@ -14,7 +14,8 @@
                 >
                     <v-icon>mdi-close</v-icon>
                 </v-btn>
-                <div id="customShortcutDialog" title="Keybinding Preference">
+                <v-text-field label="Search Command" v-model="search"></v-text-field>
+                <div v-if="search && searchCommand().length" id="customShortcutDialog" title="Keybinding Preference">
                     <!-- Edit Panel -->
                     <div id="edit" tabindex="0" @keydown="updateEdit($event)">
                         <span style="font-size: 14px">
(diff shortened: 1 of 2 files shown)


Arnabdaz:
fixed

Issue #54, opened 2023-01-07
Edit icon on hover in Custom Shortcuts


devartstar:
**Describe the bug**
Hovering on custom shortcut must show an edit(`pencil`) icon.

**To Reproduce**
1. Go to Navbar -> Tools
2. Click on Custom Shortcut
3. Hover on any of them
4. Compare with [www.circuitverse.org](www.circuitverse.org)

**Screenshots**
<img width="719" alt="Screenshot 2023-01-07 at 5 32 51 PM" src="https://user-images.githubusercontent.com/61665451/211151183-2a13f7ab-cbcf-4319-8de6-c0d690005d05.png">

**Device Information [optional]:**

**Are you working on this issue? (Yes/No)**
No

palashb01:
Hey @devartstar , I would love to work on this, could you please assign this Issue to me, Thanks.

Shu12388y:
Hi, shubham here can you assign this issue to me.  I would love to work on this issue

palashb01:
Hello @Shu12388y , i have already raised a PR and it is under review you can checkout other issues,. Thanks.

Krupalll20:
heyyy @palashb01 is this issue still open to contribute?

Adarsh9306:
Heyy @devartstar if the issue is still open , can I work on this issue? please assign it to me.

souravgope:
i want to work on this issue please assgin me

PR #58, opened 2023-01-07
add: Pencil icon on hover in custom shortcut panel.

palashb01:
Added a pencil-outline icon on hover in custom shortcuts panel.
Fixes: #54

- added a pencil-outline icon

circuitverse

### DIFF
--- a/src/components/DialogBox/CustomShortcut.vue
+++ b/src/components/DialogBox/CustomShortcut.vue
@@ -40,7 +40,7 @@
                             v-for="(keyOption, index) in keyOptions"
                             :key="index"
                         >
-                            <span id="edit-icon"></span>
+                            <span id="edit-icon"><v-icon>small mdi-pencil-outline</v-icon></span>
                             <div>
                                 <span id="command">{{ keyOption[0] }}</span>
                                 <span id="keyword"></span>


devartstar:
@palashb01 can you use the same icon as used in the current simulator to maintain consistency?

palashb01:
> @palashb01 can you use the same icon as used in the current simulator to maintain consistency?

Since you used MDI with vuetify for the icons, I implemented this with the same library but there is no icon similar or exact to the one being used in the main simulator, will change the library to find the exact icon

devartstar:
or you can use the edit_icon png which is already present.
`.../simulator/src/img/edit_icon.png`

palashb01:
@devartstar I have made the changes, kindly review the PR, Thanks.

tachyons (changes requested):
Not working, you can check the deploy preview to verify

palashb01:
> Not working, you can check the deploy preview to verify

it's working locally not sure why the preview is showing the changes, could you please verify it once by running the instance locally, Thanks.

**Demo:**
circuit

tachyons:
If it is not working in preview, it won't work in production as well. It is likely that file path format you specified is not compatible with the loader

devartstar:
@palashb01 its not working because of the way you are adding URL to background -> `url(src/assets/img/edit_icon.png)`
after building, the location of the files changes hence it wont be able to find the image.
image
Try adding relative path or importing the image.
you can run the command `npm run build` and `npm run preview` to build and check.

Arnabdaz:
fixed by #112

PR #212, opened 2023-12-18
fix: remove 'overflow' property in user-field class

Rajveer07:
### Before 
Screenshot 2023-12-18 210101
### After 
Screenshot 2023-12-18 210216


Arnabdaz (approved):
cc @vedant-jain03

PR #227, opened 2023-12-26
feat: fixed printing bool tables

### LINKED ISSUE #198: Printing Combinational Analysis - Faulty UI
**Describe the bug**
While printing in the combinational analysis part, visibility of the scrollbar was a problem - it had a scrollbar on the side while printing.
Also, the max width was set - it could not be variable.

**To Reproduce**
1. Go to Tools
2. Click on Combinational Analysis
3. Create a table. Preferably with somewhat larger variable size. (Eg. ABCDE)
4. Click on Print Table.
5. View the table. Click on cancel in the pop-up that appears.
6. The browser pop-up window also had a max-width set, which was hindering views.

**Expected behavior**
Users would expect the whole table to show on the page. There seems to be no reason why the table should have constrained height.

**Screenshots**
image

**Device Information [optional]:**
- OS: Windows
- Browser: Brave [Chromium]
- Version 1.61.101

**Are you working on this issue? (Yes/No)**
Yes.

Yarlord:
Fixes #

The print tables in boolean tables couldn't generate the whole pdf, instead only a portion of it was visible. I fixed the small issue in UI part, so that the whole truth table could be generated in the PDF format (tldr; scroll issue fixed in print table).
resolve #198 

#### Before
image
Here, only one page was visible. 

#### After
image
Here, all the pages and cells in the table are visible.

### DIFF
--- a/src/components/DialogBox/CombinationalAnalysis.vue
+++ b/src/components/DialogBox/CombinationalAnalysis.vue
@@ -699,7 +699,7 @@ function printBooleanTable() {
         table, th, td {border: solid 1px #DDD;border-collapse: 0;}
         tbody {padding: 2px 3px;text-align: center;} 
         </style>`.replace(/\n/g, "")
-    var win = window.open('', '', 'height=1d 400,width=700')
+    var win = window.open('', '', 'height=700,width=700')
     var htmlBody = `
                        <html><head>\
                        <title>Boolean Logic Table</title>\
--- a/src/components/DialogBox/CombinationalAnalysis.vue
+++ b/src/components/DialogBox/CombinationalAnalysis.vue
@@ -8,7 +8,7 @@
         :is-persistent="true"
         :table-header="tableHeader"
         :table-body="tableBody"
-        message-text="BooleanLogicTable"
+        message-text="Boolean Logic Table"
         @button-click="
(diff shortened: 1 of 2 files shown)


src/components/DialogBox/CombinationalAnalysis.vue
```diff
@@ -690,10 +690,16 @@ function generateCircuit() {
...
+
     var style =
-        '<style> table {font: 20px Calibri;} table, th, td {border: solid 1px #DDD;border-collapse: collapse;} padding: 2px 3px;text-align: center;} </style>'
-    var win = window.open('', '', 'height=700,width=700')
+        `<style>
+        table {font: 40px Calibri;}
+        table, th, td {border: solid 1px #DDD;border-collapse: 0;}
+        tbody {padding: 2px 3px;text-align: center;} 
+        </style>`.replace(/\n/g, "")
+    var win = window.open('', '', 'height=1d 400,width=700')
```
Arnabdaz, line 702: keep the height 700
```suggestion
    var win = window.open('', '', 'height=700,width=700')
```

Arnabdaz (changes requested):

Arnabdaz (approved):

PR #245, opened 2024-01-03
Removing jQuery from and cleaning up HexBinDec.vue

JoshVarga:
#### Describe the changes you have made in this PR - Remove the use of jQuery in the HexBinDec.vue file and make it more Vue-like.

### DIFF
--- a/src/components/DialogBox/HexBinDec.vue
+++ b/src/components/DialogBox/HexBinDec.vue
@@ -1,47 +1,47 @@
 <template>
-  <v-dialog
-      v-model="SimulatorState.dialogBox.hex_bin_dec_converter_dialog"
-      :persistent="false"
-  >
-      <v-card class="messageBoxContent">
-          <v-card-text>
-              <p class="dialogHeader">Hex-Bin-Dec Converter</p>
-              <v-btn
-                  size="x-small"
-                  icon
-                  class="dialogClose"
-                  @click="
-                      SimulatorState.dialogBox.hex_bin_dec_converter_dialog = false
-                  "
-              >
-                  <v-icon>mdi-close</v-icon>
-              </v-btn>
-              <div
-              v-for="(value, index) in Object.entries(inputArr)"
-                  id="bitconverterprompt"
-                  :key="value[0]"
-                  title="Dec-Bin-Hex-Converter"
-              >
-                  <label>{{ value[1].label }}</label>
-                  <br />
-                  <input
-                      :id="value[0]"
-                      type="text"
-                      :value="value[1].val"
-                      :label="value[1].label"
-                      name="text1"
-                      @keyup="(payload) => converter(payload)"
-                  />
-                  <br /><br />
-              </div>
-          </v-card-text>
-          <v-card-actions>
-              <v-btn class="messageBtn" block @click="setBaseValues(0)">
-                  Reset
-              </v-btn>
-          </v-card-actions>
-      </v-card>
-  </v-dialog>
+    <v-dialog
+        v-model="SimulatorState.dialogBox.hex_bin_dec_converter_dialog"
+        :persistent="false"
+    >
+        <v-card class="messageBoxContent">
+            <v-card-text>
+                <p class="dialogHeader">Hex-Bin-Dec Converter</p>
+                <v-btn
+                    size="x-small"
+                    icon
+                    class="dialogClose"
+                    @click="
+                        SimulatorState.dialogBox.hex_bin_dec_converter_dialog = false
+                    "
+                >
+                    <v-icon>mdi-close</v-icon>
+                </v-btn>
+                <div
+                    v-for="(value, index) in Object.entries(inputArr)"
+                    id="bitconverterprompt"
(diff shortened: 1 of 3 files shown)


src/components/DialogBox/HexBinDec.vue
```diff
@@ -1,136 +1,134 @@
...
 
-function decimalConvertor() {
-    var x = parseInt($('#decimalInput').val(), 10)
-    setBaseValues(x)
+function setBaseValues(x: number) {
+  if (isNaN(x)) {
+    return;
+  }
+  inputArr.value.binaryInput.val = '0b' + x.toString(2);
+  inputArr.value.bcdInput.val = parseInt(x.toString(10), 16).toString(2);
```
Arnabdaz, line 86: maybe we update the bcd function to always generate 4 binary digits 
wdyt ?

```
function convertToBCD(number) {
    let digits = number.toString().split('');
    let bcdOfDigits = digits.map(function(digit) {
        return parseInt(digit).toString(2).padStart(4, '0');
    });
    return bcdOfDigits.join('');
}
```
this would make sure to format the bcd correctly
like for `2` instead of `10` it would show `0010`

Arnabdaz (changes requested):

src/components/DialogBox/HexBinDec.vue
```diff
@@ -1,136 +1,134 @@
...
-                </div>
-            </v-card-text>
-            <v-card-actions>
-                <v-btn class="messageBtn" block @click="setBaseValues(0)">
-                    Reset
-                </v-btn>
-            </v-card-actions>
-        </v-card>
-    </v-dialog>
+  <v-dialog v-model="SimulatorState.dialogBox.hex_bin_dec_converter_dialog" :persistent="false">
```
Arnabdaz, line 2: can you please use the current format of template with prettier instead of using volar format ?
as it is hard to review changes

Arnabdaz (changes requested):

src/components/DialogBox/HexBinDec.vue
```diff
@@ -47,90 +47,114 @@
...
-    {
-        inputId: 'binaryInput',
+    binaryInput: {
         val: '0b10000',
         label: 'Binary value',
     },
-    {
-        inputId: 'bcdInput',
+    bcdInput: {
         val: '10110',
```
Arnabdaz, line 62:
```suggestion
        val: '00010110',
```

src/components/DialogBox/HexBinDec.vue
```diff
@@ -47,90 +47,114 @@
...
+    } else {
+        x = parseInt(input, 2)
+    }
     setBaseValues(x)
 }
 
-function bcdConvertor() {
-    var input = $('#bcdInput').val()
+function bcdConverter(input: string) {
     var num = 0
```
Arnabdaz, line 135: let's use `let` & `const` instead of `var`

src/simulator/src/utils.js
```diff
@@ -269,50 +248,6 @@ export function parseNumber(num) {
     return parseInt(num)
 }
 
-export function setupBitConvertor() {
-    console.log('check bit convertor')
-    $('#decimalInput').on('keyup', function () {
```
Arnabdaz, line 274: we lost the functionality of rejecting wrong values 
eg: in binary if we add anything else then `0`,`1` it gets removed 

[Screencast from 2024-01-04 10-25-27.webm](https://github.com/CircuitVerse/cv-frontend-vue/assets/96580571/66dcec26-3ce3-42a7-8b27-463329b92fee)

Arnabdaz (changes requested):

src/simulator/src/utils.js
Arnabdaz: we can use the old method of `parseInt` or better we can use `regex` and have the functionality

Arnabdaz:
```js
function converter(e: KeyboardEvent) {
    const target = <HTMLInputElement>e.target!
    let value = target.value;

    // Regular expressions for validating input
    const regexBinary = /[^01]/g;
    const regexOctal = /[^0-7]/g;
    const regexDecimal = /[^0-9]/g;
    const regexHex = /[^0-9A-Fa-f]/g;

    switch (target.id) {
        case 'decimalInput':
            value = value.replace(regexDecimal, '');
            decimalConverter(value);
            break;
        case 'binaryInput':
            value = value.replace(regexBinary, '');
            binaryConverter(value);
            break;
        case 'bcdInput':
            value = value.replace(regexBinary, '');
            bcdConverter(value);
            break;
        case 'octalInput':
            value = value.replace(regexOctal, '');
            octalConverter(value);
            break;
        case 'hexInput':
            value = value.replace(regexHex, '');
            hexConverter(value);
            break;
    }

    // Update the input field with the cleaned value
    target.value = value;
}
```

use this

Arnabdaz (approved):
LGTM nice work :+1: 
cc @tachyons @devartstar

PR #326, opened 2024-06-23
Remove unused files, unused imports, commented out code, and update NPM packages

JoshVarga:
General cleanup


src/simulator/src/modules/Ground.js
```diff
@@ -54,21 +50,6 @@ export default class Ground extends CircuitElement {
         simulationArea.simulationQueue.add(this.output1)
     }
 
-    /**
-     * @memberof Ground
-     * fn to create save Json Data of object
-     * @return {JSON}
-     */
-    customSave() {
```
gr455, line 62: I believe this definition of customSave is being used currently. The earlier one is redundant (the Ground constructor does not take the direction, and Ground does not have a state).

src/simulator/src/modules/Output.js
```diff
@@ -59,15 +56,6 @@ export default class Output extends CircuitElement {
         this.inp1 = new Node(this.bitWidth * 10, 0, 0, this)
     }
 
-    /**
-     * @memberof Output
-     * function to generate verilog for output
-     * @return {string}
-     */
-    generateVerilog() {
```
gr455, line 67: This function generates verilog for the output node, removing it will break that feature.

src/simulator/src/modules/Splitter.js
```diff
@@ -183,53 +178,12 @@ export default class Splitter extends CircuitElement {
         this.prevInpValue = undefined
     }
 
-    /**
-     * @memberof Splitter
-     * fn to process verilog of the element
-     * @return {JSON}
-     */
-    processVerilog() {
```
gr455, line 191: Why was this removed?

src/simulator/src/modules/Ground.js
JoshVarga: Currently Ground.js has customSave() defined twice, in javascript the second one wins

gr455: Yep, we would want the second one to keep winning since that is the correct implementation. So I think it would be better to remove the [first implementation](https://github.com/CircuitVerse/cv-frontend-vue/blob/d0787734b8183f83f67d11c31c6f888c7c8f8058/src/simulator/src/modules/Ground.js#L31)

src/simulator/src/modules/Splitter.js
JoshVarga: Here too generateVerilog is defined twice for the same object

src/simulator/src/modules/Output.js
gr455: Gotcha. This is good then. Thanks

src/simulator/src/modules/Splitter.js
gr455: Got it.

gr455 (approved):
LGTM

gr455:
@tachyons Can you merge this. I don't seem to have write access. (also please give me write access too)

PR #333, opened 2024-07-06
fix: update project name in navbar 

hackwhiz-8:
Fixes #

In the navbar there is half portion is hidden because the width issue so i just basically made changes to the styling of that box. and now its perfectly running .
I have also shared the snapshot of fixed navbar .

**before**

circuitVerseNavbar

**After** 

Screenshot 2024-07-06 224021


aryanndwi123:
Stale PR!!! Closing as this is already solved in the previous PRs

PR #395, opened 2024-12-09
Changes behavior for simultaneous async preset and reset high.

G-K-404:
The changes in the code help better to moderate the state behavior during Race condition.

Fixes # When both async preset and reset are high, the state is set to previous state instead of toggling to high.

image
above is changes for dff, similar changes have been all made in codes of all flip flops.

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of asynchronous reset and preset conditions across various flip-flop classes (D, JK, SR, T).
	- Improved robustness against race conditions in state management.

- **Bug Fixes**
	- Refined logic for state updates to ensure correct behavior when both preset and reset are active.

- **Documentation**
	- Updated comments for clarity on the new behavior of flip-flops under specific conditions.

### DIFF
--- a/src/simulator/src/sequential/DflipFlop.js
+++ b/src/simulator/src/sequential/DflipFlop.js
@@ -61,8 +61,10 @@ export default class DflipFlop extends CircuitElement {
      * and input of the clock. We flip the bits to find qInvOutput
      */
     resolve() {
-        if (this.reset.value == 1) {
-            this.masterState = this.slaveState = this.preset.value || 0
+        if (this.reset.value == 1 && this.preset.value == 0) {
+            this.masterState = this.slaveState = 0;
+        } else if(this.preset.value == 1 && this.reset.value == 0){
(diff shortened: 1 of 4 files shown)


ThatDeparted2061 (commented):
The code here 
this.masterState = this.slaveState = this.preset.value || 0
gives precedence to preset over reset
In circuit designing it is avoided to have both reset(get system to 0) and preset(get system to 1) high but here to counter this the precedence is given to preset .
This is not like SR latch since the S R inputs in SR latch are a part of the steps for changing the output as per the input values but reset and preset are the master options to change the output to 0 or 1 irrespective of the input, so they dont have anything  to do with the previous state.
### The changes you have made do not answer to the condition when reset and preset both are high, and this can lead to an undefined state.
This undefined state is answered well by the previous code

ThatDeparted2061:
For the above mentioned reasons @G-K-404 kindly close the PR .

PR #619, opened 2025-06-29
Made the verilog module consistent with the Circuit simulation logic

### LINKED ISSUE #561: Fix the errors in the Verilog module of the circuit elements

### Describe the bug

Check and fix the errors in the Verilog module
of the circuit elements and fix the bugs found

1. Many circuit elements in CircuitVerse have bugs in their Verilog module definitions and are not consistent with their simulation model.
For example, all flip-flop modules include pre (preset) in their parameter list, but:
- Pre is not implemented in the Verilog code,
- Even though it exists in the actual circuit element's logic implementation.
3. Other circuit components show similar inconsistencies and bugs between:
- The Verilog module code, and
- Their logic implementation within the CircuitVerse simulator.

092vk:
Fixes #561 

1. Added the logic for pre parameter in the verilog module
2. According to simulation logic pre is not a control signal but a register which sets the q or output


niladrix719 (approved):
lgtm

PR #623, opened 2025-06-30
feat(gsoc2025): add pull resistor component

### LINKED ISSUE #567: Feat: Shift Registers Component
Shift registers are sequential logic components that store and move data in a specific direction on each clock cycle.

We can allow users to switch between various modes is single input or multiple input through properties panel. There will be two modes Load/Shift. When load signal is enabled, data will be loaded otherwise shifted.

### LINKED ISSUE #584: Feat: Pull-down Resistor
A pull-down resistor is a resistor connected between a digital input pin and ground (GND) to ensure the pin stays at a default LOW (0) logic level when no active signal is present. It prevents the input from floating, which can cause unpredictable behavior in digital circuits.

Nihal4777:
Fixes #584 

Added pull resistor component & modified Properties panel to support drop down type of component specific properties.

Comparator


v0/src/simulator/src/modules/PullResistor.js
```diff
@@ -0,0 +1,127 @@
...
+        this.inp = new Node(0, -10, 0, this, 1, 'inp')
+    }
+
+     /**
+     * Creates JSON data for saving the component state
+     * @returns {Object} Save data including constructor parameters and nodes
+     */
+    customSave() {
+        const data = {
+            constructorParamaters: [this.direction, this.bitWidth],
```
Asrani-Aman, line 36: @Nihal4777 why this.bitwidth is here?

I am not able to get it why pullRegisete need to store the bit width?

Asrani-Aman: I think we should store the PullDirection also ?

v0/src/simulator/src/modules/PullResistor.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Changes the pull direction property and redraws the component
+     * @param {string} pullDirection - New pull direction
+     */
+    changePullDirection(pullDirection) {
+        if (pullDirection !== undefined && this.pullDirection !== pullDirection) {
+            this.pullDirection = pullDirection;
+            var obj = new PullResistor(this.x, this.y, this.scope, this.dir, this.pullDirection)
```
Asrani-Aman, line 98: why we are creating new PullRegister, can't we directly update the existing ?

v0/src/simulator/src/modules/PullResistor.js
```diff
@@ -0,0 +1,127 @@
...
+ */
+export default class PullResistor extends CircuitElement {
+    constructor(x, y, scope = globalScope, dir = 'RIGHT', pullDirection = "Down") {
+        super(x, y, scope, dir, 1)
+
+        this.rectangleObject = false;
+        this.fixedBitWidth = true;
+        this.setDimensions(10, 30)
+        this.pullDirection = pullDirection ?? "Down";
+        this.inp = new Node(0, -10, 0, this, 1, 'inp')
```
Asrani-Aman, line 27: this node should be of type2 intermediate, right?

Nihal4777: No, Type 2 is for those two wire connecting points.

v0/src/simulator/src/modules/PullResistor.js
Nihal4777: Actually, In other components new object is created on mutable properties change

v0/src/simulator/src/modules/PullResistor.js
Nihal4777: Added

Asrani-Aman:
@Nihal4777 description image is of comaparator

v0/src/simulator/src/modules/PullResistor.js
Asrani-Aman: I didn't get your point,
but while testing doing type 2 is not wroking

Can we discuss more on this on a meet ?

Issue #260, opened 2024-01-16
Bug (UI): When ElementsPanel is moved to overlap the Navbar, it becomes locked there and loses the ability to be dragged.


Chandrachur67:
**Describe the bug**
The draggable ElementsPanel can be moved to overlap with the Navbar component and if we leave that there then we lose the functionality to move the ElementsPanel further because the Navbar appears on top of the ElementsPanel Component.

**To Reproduce**
1. Go to [cv-frontend-vue](https://circuitverse.netlify.app/simulatorvue)
2. Move the ElementsPanel such that it overlaps the Navbar at the top of the screen.
3. Leave it there
4. Now try to move the ElementsPanel component
5. Error: You won't be able to move it.

**Expected behavior**
I can think of two possible solutions.
1. Making the ElementsPanel appear on top of Navbar making it always free to move.
2. Restricting the ElementsPanel to overlap with the Navbar

**Screenshots**

*This is movable*
![ss_1 copy](https://github.com/CircuitVerse/cv-frontend-vue/assets/97349505/4a320961-8513-4fe4-8345-042e6a969eec)
*This is not movable*
![ss_2 copy](https://github.com/CircuitVerse/cv-frontend-vue/assets/97349505/ba8aed77-0301-45b5-b127-446754ad45cd)

**Device Information [optional]:**
- OS: macOS
- Browser Brave
- Version 1.61.109

**Additional context**
I think need to think of one more edge case i.e. when Navbar options are expanded. Considering this case the first possible solution seems better to me.

**Are you working on this issue? (Yes/No)**
Yes

Arnabdaz:
@Chandrachur67 you can work on the issue check now the tabsbar must also be having the same issue after merging #223 fix that also ... mostly its a issue of z-index.

Chandrachur67:
Thanks for the clarification. The solution is almost ready. I will raise a pr very soon.

29deepanshutyagi:
hi , i want to work on this issue , as there is an inactivity on this issue from various months ,so kindly assign me @Arnabdaz

SrisharanVS:
@Arnabdaz 
Hey, I have fixed this.. Can you assign me this, I will create a PR soon..

rajsingh112005:
@Arnabdaz  
hey, I Have completed my work on this issue plz assign it to me

SrisharanVS:
@rajsingh112005 Many did, I think this is ded bro...:)

[...]

PR #691, opened 2025-12-09
fix(ui): prevent ElementsPanel from being dragged under Navbar (#260)

akritah:
**_Fix: Prevent ElementsPanel from overlapping Navbar when dragged (#260)_**
**Description**
This PR fixes the issue where the ElementsPanel (and other draggable panels) could be dragged over the Navbar, causing visual overlap and poor UX.

**Changes Made:**

- Modified updatePosition() function in drag.ts to add navbar collision detection
- Dynamic navbar height detection using getBoundingClientRect() instead of hardcoded values
- Constraint logic: Prevents panel's top position from going above the navbar's bottom edge
- Applied consistently across all three versions: src, v0, and v1
- Added unit tests in drag.spec.js for navbar overlap prevention

**The fix works by:**
- Querying the navbar element .navbar.header during each drag operation
- Getting its runtime bottom position via getBoundingClientRect()
- Calculating the new panel position and constraining it to stay below the navbar
- Preserving original drag UX and event cleanup behavior

**Testing**

- Panel cannot be dragged above navbar
- Panel moves freely when not near navbar
- Works correctly when navbar doesn't exist (graceful fallback)
- Uses runtime navbar height (no hardcoded values)
- Original drag behavior preserved
- Unit tests added and passing

https://github.com/user-attachments/assets/a871c6a8-16c2-4154-8f12-9caa7c325450

## Release Notes

* **Bug Fixes**
  * Fixed panel dragging behavior to prevent overlap with the fixed navbar when panels are moved.
  * Improved z-index stacking for interactive UI elements to maintain proper layering.

* **Tests**
  * Added test suite for panel dragging with navbar interaction verification.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/components/Panels/ElementsPanel/ElementsPanel.vue
+++ b/src/components/Panels/ElementsPanel/ElementsPanel.vue
@@ -220,6 +220,10 @@ function getTooltipText(elementName: string) {
 </script>
 
 <style>
+.elementPanel {
+    z-index: 120; /* sit above navbar (z-index 100) while remaining draggable */
+}
+
 .v-expansion-panel-title {
     min-height: 36px;
 }
--- a/src/components/Panels/ElementsPanel/ElementsPanel.vue
+++ b/src/components/Panels/ElementsPanel/ElementsPanel.vue
@@ -220,10 +220,6 @@ function getTooltipText(elementName: string) {
 </script>
 
 <style>
-.elementPanel {
-    z-index: 120; /* sit above navbar (z-index 100) while remaining draggable */
-}
-
 .v-expansion-panel-title {
     min-height: 36px;
 }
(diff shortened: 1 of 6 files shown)


Nihal4777 (commented):
Thanks @akritah for raising the PR. We would suggest to use z-index to handle element panels properly as discussed in #260 same like the old simulator at circuitverse.org/simulator. Let me know if you need any help or clarification.

akritah:
I’ve updated the fix to fully align with the old simulator behavior by handling the ElementsPanel layering via z-index (and positioning) so it correctly overlays the navbar and remains draggable.

Thanks for the guidance @Nihal4777 ! Happy to make any further adjustments if needed.

Nihal4777:
Hi @akritah, Thanks for the adjustments. Actually, there is no need to modify drag.js. you just have to apply z-index in css such the all panels maintain their relative layering just like the old simulator.

Harkeerat24:
The issue has been resolved, no need for this PR

PR #859, opened 2026-01-24
fix: Correct Euclidean distance calculation formula

### LINKED ISSUE #860: Bug: Incorrect Euclidean distance calculation in distance() function
## Description
The `distance()` function in `src/simulator/src/utils.ts` has a critical mathematical error that produces incorrect distance calculations.

## Current Implementation
```typescript
export function distance(x1: number, y1: number, x2: number, y2: number) {
    return Math.sqrt((x2 - x1) ** 2) + (y2 - y1) ** 2
}
```

## The Problem
The second squared term is placed outside the `Math.sqrt()` call, resulting in:
- √((x₂-x₁)²) + (y₂-y₁)² (incorrect)

Instead of the correct Euclidean distance formula:
- √((x₂-x₁)² + (y₂-y₁)²) (correct)

## Impact
This bug affects any feature that depends on accurate distance calculations:
- Element positioning and snapping
- Wire routing and collision detection
- Selection bounds checking
- Spatial operations throughout the simulator

## Expected Behavior
The function should return the correct Euclidean distance between two points using the standard mathematical formula.

## Solution
Move the addition operation and second squared term inside the `Math.sqrt()` call:
```typescript
export function distance(x1: number, y1: number, x2: number, y2: number) {
    return Math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2)
}
```

## PR Reference
Fixed in PR #859

s1dhu98:
## Issue Description
Fixed a critical mathematical error in the distance() function in src/simulator/src/utils.ts

## The Bug
The distance calculation had incorrect parentheses:
```
Before: Math.sqrt((x2 - x1) ** 2) + (y2 - y1) ** 2
After:  Math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2)
```

The second squared term was outside the square root function.

## Impact
This bug affected any feature that depends on accurate distance calculations including element positioning, wire routing, collision detection, and selection bounds checking.

## Fix
Moved the addition and second squared term inside the sqrt() call to correctly implement the Euclidean distance formula: d = √((x₂-x₁)² + (y₂-y₁)²)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the distance calculation used by the simulator so Euclidean distance is computed properly, improving accuracy of all distance-dependent measurements and behaviors.
  * No public APIs or function signatures were changed; this is an internal accuracy fix with no interface impact.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/simulator/src/utils.ts
+++ b/src/simulator/src/utils.ts
@@ -71,7 +71,7 @@ export function showMessage(mes: string) {
 }
 
 export function distance(x1: number, y1: number, x2: number, y2: number) {
-    return Math.sqrt((x2 - x1) ** 2) + (y2 - y1) ** 2
+    return Math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2)
 }
 
 /**
--- a/v1/src/simulator/src/utils.ts
+++ b/v1/src/simulator/src/utils.ts
@@ -71,7 +71,7 @@ export function showMessage(mes: string) {
 }
 
 export function distance(x1: number, y1: number, x2: number, y2: number) {
-    return Math.sqrt((x2 - x1) ** 2) + (y2 - y1) ** 2
+    return Math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2)
 }
 
 /**


s1dhu98:
@tachyons 

## Impact Analysis

The bugis in the **distance calculation formula**. Currently it calculates:
```
distance = √((x₂-x₁)²) + (y₂-y₁)²
```

But should calculate:
```
distance = √((x₂-x₁)² + (y₂-y₁)²)
```

## Practical Example
Consider two points: (0, 0) and (3, 4)
- **Current (buggy) formula**: √(9) + 16 = 3 + 16 = **19** ❌
- **Correct formula**: √(9 + 16) = √25 = **5** ✓

## Where It Causes Issues
1. **Selection bounds** - Elements far away are incorrectly marked as "close" or vice versa
2. **Wire snapping** - Connections snap to wrong nodes 
3. **Collision detection** - Elements may incorrectly collide or fail to detect collision
4. **Auto-alignment** - Element positioning is inaccurate
5. **Visual operations** - Pan/zoom and spatial queries use incorrect distances

## Why It Works Despite the Bug
The simulator still appears to work because the distance function is mostly used for **comparisons** ("is A closer than B?") rather than absolute values. Since both buggy and correct values increase monotonically, relative ordering is preserved in most cases. However, this breaks down for critical spatial operations.

This is a subtle but critical bug that affects the geometric integrity of the simulator.

PR #893, opened 2026-01-29
fix: right-click delete in lower part of board (#829)

### LINKED ISSUE #829: 🐞Bug: Not gate and wire segment cannot be deleted after aborted connection

### Describe the bug

Description

When starting a wire connection and not completing it (leaving a dangling wire segment), the dangling red wire remains on the canvas and cannot be removed using normal delete actions.Same with a not gate too

### Steps to Reproduce

1)Open Vue simulator (v0)

2)Place any component (ex: NOT gate or Input)

3)Start drawing a wire from a pin (click and drag)

4)Release the wire without connecting to another pin

5)A small dangling red wire segment appears

6)Try deleting it (Select tool + Delete/Backspace / Esc / Right click)

### Expected Behavior

Dangling wire should be removed when:

pressing Esc (cancel connection), or

selecting and deleting it, or

undoing the action

### Screenshots

Image
Image

Yes

None

None

### Device Information

```markdown
Simulator version: Vue simulator v0

Platform: Windows 11 (Chrome)

Reproducible consistently

Screenshot attached (showing red dangling wire)
```

### Additional Context

_No response_

Yes

Nakshatra480:
## What
Right click → Delete (and Copy/Cut) was not working when the component was in the lower part of the canvas. It worked in the upper part and keyboard Delete still worked everywhere.

## Why
Selection for the context menu actions was coming from the mousedown that opened the menu. In the lower part of the board that selection wasn’t reliable, so `lastSelected` was wrong or missing when the user chose Delete.

## How
Before showing the context menu we now sync selection to the right click position: convert that position to canvas/simulation coords (same as `panStart` in listeners), set the relevant `simulationArea` mouse fields, and run `update(globalScope, true)` so the element under the cursor becomes `lastSelected`. That way Delete/Copy/Cut always target the element the user right-clicked on, no matter where it is on the canvas.

Closes #829

## Summary by CodeRabbit

* **Bug Fixes**
  * Context menu now syncs with current selection and mouse state so menus reliably appear in the correct spot.
  * Canvas/selection coordinates are refreshed before showing the menu, reducing stale or misaligned menus.

* **Style**
  * Context menu visual layout and sizing updated for more stable positioning and improved appearance.

* **Behavior Changes**
  * Multi-selection deletion logic adjusted, affecting which items can be removed in batch actions.

### DIFF
--- a/src/components/ContextMenu/ContextMenu.vue
+++ b/src/components/ContextMenu/ContextMenu.vue
@@ -1,5 +1,5 @@
 <template>
-    <div id="contextMenu" oncontextmenu="return false;">
+    <div id="contextMenu" oncontextmenu="return false;" @mousedown.stop>
         <ul>
             <li
                 v-for="(menuOption, index) in contextMenuOptions"
--- a/src/simulator/src/ux.js
+++ b/src/simulator/src/ux.js
@@ -73,10 +73,10 @@ function syncSelectionToContextMenuPosition() {
     simulationArea.mouseYf = (rawY - globalScope.oy) / globalScope.scale
     simulationArea.mouseX = simulationArea.mouseDownX
     simulationArea.mouseY = simulationArea.mouseDownY
-    var wasMouseDown = simulationArea.mouseDown
+    var prevMouseDown = simulationArea.mouseDown
     simulationArea.mouseDown = true
     update(globalScope, true)
-    simulationArea.mouseDown = wasMouseDown
+    simulationArea.mouseDown = prevMouseDown
 }
 
 /**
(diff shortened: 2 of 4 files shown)


Nihal4777 (changes requested):
I noticed that when the context menu opens in the lower part of the canvas, it renders upward. However, the outside-click detection logic assumes the menu always opens downward (using ctxPos.y to ctxPos.y + offsetHeight). Because of this, clicks on menu items are incorrectly treated as clicks outside the menu, causing it to close before the item click handler runs.

Nakshatra480:
I have implemented the fix by changing wasMouseDown to prevMouseDown in both ux.js files. This ensures proper mouse state preservation during right click operations, preventing the persistent dragging state. The changes maintain consistency and resolve the lower board area issue.

Nakshatra480:
Hi @Nihal4777 pls review this PR, changes are properly working.

https://github.com/user-attachments/assets/32a43b81-2956-4215-886c-18c198237819

thanks for you time

PR #916, opened 2026-02-02
Fix Euclidean distance calculation in pinch to zoom logic

### LINKED ISSUE #860: Bug: Incorrect Euclidean distance calculation in distance() function

Nakshatra480:
This PR fixes a bug in the pinch to zoom gesture handler where the Euclidean distance was being calculated incorrectly. The `Math.sqrt` function was being passed two separate arguments instead of the sum of squared differences, which caused inaccurate zoom behavior.

**Key Changes:**
- Corrected the distance formula in `src/simulator/src/listeners.js` and the v1 equivalent.
- Added a new unit test `pinchZoom.spec.js` to verify the fix and prevent regressions.
- Updated `vitestSetup.ts` with a lightweight `localStorage` mock to resolve test environment failures.

**Related Issue:**
Fixes #860

**Verification:**
- Added a specific test case that confirms the scale increases correctly when the touch distance expands.
- Verified that all existing tests pass with the updated setup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected pinch-zoom distance calculation so pinch gestures produce accurate, consistent zoom behavior.

* **Tests**
  * Added automated pinch-zoom tests verifying scale progression across gestures.
  * Improved test setup with safer global mocks and an in-memory localStorage that is cleared between tests to stabilize runs.

### DIFF
--- a/src/simulator/spec/pinchZoom.spec.js
+++ b/src/simulator/spec/pinchZoom.spec.js
@@ -1,0 +1,31 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { pinchZoom } from '../src/listeners';
+
+vi.mock('../src/simulationArea', () => ({ simulationArea: { canvas: { getBoundingClientRect: vi.fn(() => ({ left: 0, top: 0, width: 100, height: 100 })) } } }));
+vi.mock('../src/engine', () => ({ gridUpdateSet: vi.fn(), scheduleUpdate: vi.fn(), updateSimulationSet: vi.fn(), updatePositionSet: vi.fn(), updateCanvasSet: vi.fn(), wireToBeCheckedSet: vi.fn(), errorDetectedSet: vi.fn() }));
+vi.mock('../src/layoutMode', () => ({ layoutModeGet: vi.fn(), tempBuffer: {}, layoutUpdate: vi.fn() }));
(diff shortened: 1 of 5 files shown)


Nakshatra480:
Hi @Nihal4777 review this PR and merge if possible.
thanks for your time

PR #1049, opened 2026-05-09
fix: fix element panel overlapping #260

### LINKED ISSUE #260: Bug (UI): When ElementsPanel is moved to overlap the Navbar, it becomes locked there and loses the ability to be dragged.

AdityaOP007:
Screenshot 2026-05-09 222558
Screenshot 2026-05-09 222608

Fixed the issue #260 

Fixes Issue #260: UI bug where the ElementsPanel (and other draggable panels) becomes locked and loses drag ability when moved to overlap the Navbar.

Cause of the Bug
The bug was caused by a z-index stacking context issue. The Navbar was statically set to z-index: 100. The draggable panels had a base z-index: 70, and the interactjs dragging logic only temporarily elevated the active panel to z-index: 99. Because 99 is still less than 100, moving a panel's header underneath the Navbar caused the Navbar to overlay the panel and intercept all pointer events, effectively "locking" it in place so it couldn't be clicked or dragged out.

Changes I made
To fix this, the z-index hierarchy of the draggable panels was elevated so that they always remain above the Navbar.

src/simulator/src/drag.ts
Updated the mousedown.dragging event listener. Now, when a panel is clicked, it elevates to z-index: 102, while all other inactive panels are pushed to z-index: 101.
src/components/Navbar/Navbar.css & src/styles/css/main.stylesheet.css
Updated the base .draggable-panel-css class, increasing its default z-index from 70 to 101. This ensures panels safely hover over the Navbar even when stationary.
Added /* AdityaOP007 solved issue #260 */ comments above the updated z-index values in all three files to document the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the stacking order of draggable panels so non-active panels sit behind the panel being dragged, preventing visual overlap and ensuring consistent layering during drag interactions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CircuitVerse/cv-frontend-vue/pull/1049?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

Fixes #260


Harkeerat24:
The issue has been resolved, no need of this PR

PR #1090, opened 2026-06-10
fix: correct BCD input handler passing undefined variable to setBaseValues

KeerthiKumarR:
The BCD input handler inside `setupBitConvertor()` in `utils.ts`
has two bugs that break the bit convertor panel when typing in
the BCD field.

**Bug 1 — undefined variable passed to setBaseValues()**

The handler decodes BCD input and accumulates the result into
`num`, but the final line calls `setBaseValues(x)` — `x` is
never declared in this scope so it is always `undefined`. Every
keystroke in the BCD field ends up calling `setBaseValues(undefined)`,
setting all other display fields (decimal, hex, octal, binary)
to NaN instead of the correct value.

```ts
// before
return setBaseValues(x);   // x is undefined

// after
return setBaseValues(num);  // num holds the decoded value
```

**Bug 2 — string compared to number**

`input` is a string from `.val()` but was compared to number `0`.
In JavaScript `"0" !== 0` is always `true`, so the guard meant
to skip empty input never worked.

```ts
// before
if (input !== 0) {

// after
if (input !== "0") {
```

Both changes are in `src/simulator/src/utils.ts`, 2 lines total.

**PR Category**
- [x] Bug Fix

**Checklist**
- [x] Self-reviewed the code
- [x] Tested — typing in BCD field now correctly updates all fields
- [x] Follows project style guidelines

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of BCD input values.
  * Ensured normalized zero input is processed correctly and the computed value is applied consistently.


KeerthiKumarR:
CC: @aryanndwi123 for review.

Issue #1069, opened 2026-05-26
(GSoC'26) Week 2: Implement Multi-Circuit Project Export and Import


aryanndwi123:

Harkeerat24:
Title:
(GSoC'26) Week 2: Implement Multi-Circuit in canonical.ts

Description::
- Adds support for multi-circuit structures inside canonical.ts.
- Iterates through all project scopes to produce a single, deterministic JSON output.

Issue #1067, opened 2026-05-26
(GSoC'26) Week 1: Improve and fix canonical.js and importCanonical.js for all circuitElementList types.


aryanndwi123:
Description:

Implement canonical.ts : This is the file that will convert the circuit to the canonical JSON.
Implement the Weisfeiler-Lehman algorithm in canonical.ts for determinism.
Determinism should be achieved for all circuit types and all of their properties.

Harkeerat24:
Title:
(GSoC'26) Week 1: Implement canonical.ts & 1-WL algorithm for all component types

Description:
- Implement canonical.ts : This is the file that will convert the circuit to the canonical JSON.
- Implement the Weisfeiler-Lehman algorithm in canonical.ts for determinism.
- Determinism should be achieved for all circuit types and all of their properties.

PR #1095, opened 2026-06-14
feat: added subcircuit support in canonical.ts using khans algorithm

### LINKED ISSUE #1070: (GSoC'26) Week 3: Include SubCircuit support. Implement topological sorting (Kahn's Algorithm) for scope resolution.

Harkeerat24:
Fixes #1070 , #1069 and #1067

 

Added subcircuit support:

- **SubCircuit identity in WL fingerprinting** — the child scope ID is appended to the initial signature (`|${subId}`)

- **`khansAlgorithm`** — Kahn's BFS-based topological sort over the scope dependency graph. Uses a sorted queue (smallest scope ID dequeued first) so the output order is deterministic when multiple scopes are ready simultaneously. Returns `null` if a cycle is detected, which `canonicaliseProject` converts into a thrown error.

No UI Changes
---

- Fixed Hash Collisions: Previously, different subcircuits with similar properties were assigned the same hash. We now append the unique child scope ID to the signature, ensuring every unique subcircuit gets a distinct hash.
- Implemented Kahn’s Algorithm: We chose this for topological sorting because it is efficient O(V + E), easy to maintain, and produces a deterministic output order every single time.
- Cycle Detection: Kahn’s Algorithm automatically identifies invalid circular dependencies. If it finds a cycle, the system throws an error instead of processing corrupt data.

---

## Summary by CodeRabbit

* **New Features**
  * Added deterministic circuit canonicalization that produces stable, hashable representations of component connectivity, nets, exposed interface ports, and visual/layout details.
  * Added project-level canonicalization that orders scopes consistently based on subcircuit relationships and generates a stable project hash.
  * Updated subcircuit serialization to include node connectivity and constructor parameter metadata for consistent hashing.
* **Bug Fixes**
  * Improved detection and rejection of circular scope dependencies during project canonicalization.

### DIFF
--- a/v1/src/simulator/src/data/canonical.ts
+++ b/v1/src/simulator/src/data/canonical.ts
@@ -13,9 +13,9 @@ type WireNode = {
 type NodeIndexMap = Map<WireNode, number>;
 
 type ComponentSaveData = {
-  nodes: Record<string, unknown>;
-  values?: Record<string, unknown>;
-  constructorParamaters?: unknown[];
+  nodes: Record<string, WireNode | WireNode[]>;
+  values?: Record<string, string | number | boolean>;
+  constructorParamaters?: Array<string | number | boolean | Record<string, unknown>>;
 };
 
 type CVComponent = {
@@ -23,10 +23,10 @@ type CVComponent = {
   label?: string;
   bitWidth: number;
   customSave: () => ComponentSaveData;
-  direction?: unknown;
+  direction?: "LEFT" | "RIGHT" | "UP" | "DOWN";
   propagationDelay?: number;
   state?: unknown;
-  labelDirection?: unknown;
+  labelDirection?: "LEFT" | "RIGHT" | "UP" | "DOWN";
   x?: number;
   y?: number;
   [key: string]: unknown;
@@ -65,11 +65,11 @@ type ComponentDraft = {
   properties: Record<string, unknown>;
   _connections: Record<string, number>;
   _state?: unknown;
-  _labelDirection?: unknown;
+  _labelDirection?: "LEFT" | "RIGHT" | "UP" | "DOWN";
   _x?: number;
   _y?: number;
   _instance?: CVComponent;
-  _portDefs?: Record<string, unknown>;
+  _portDefs?: Record<string, WireNode | WireNode[]>;
   [key: string]: unknown;
 };
 
@@ -113,7 +113,7 @@ type SubcircuitSymbolLayout = {
 
 type CanonicalLayout = {
   [componentId: string]:
-    | { x?: number; y?: number; labelDirection?: unknown; [key: string]: unknown }
+    | { x?: number; y?: number; labelDirection?: "LEFT" | "RIGHT" | "UP" | "DOWN"; [key: string]: unknown }
     | Record<string, IntermediateNet>
     | SubcircuitSymbolLayout
     | undefined;
@@ -348,7 +348,7 @@ function buildComponentDrafts(scope: CVScope, uf: UnionFind, nodeIndexMap: NodeI
         }
       }
 
-      const defaultState =
+      const defaultState: unknown | undefined =
         statePropKey !== undefined && (comp as Record<string, unknown>)[statePropKey] !== undefined
           ? (comp as Record<string, unknown>)[statePropKey]
           : undefined;
@@ -913,7 +913,7 @@ export async function canonicaliseScope(
         comp = { ...comp, defaultState: undefined };
       }
       if (DIRECTION_BEARING.has(c.type) && Array.isArray(comp.properties?.constructorParamaters)) {
-        const params = (comp.properties.constructorParamaters as unknown[]).slice();
+        const params = comp.properties.constructorParamaters.slice();
         // Null direction (index 0) for direction-bearing components.
         if (params.length > 0) {
           params[0] = null;
@@ -926,7 +926,7 @@ export async function canonicaliseScope(
       }
       // Replace the numeric child scope ID in SubCircuit constructorParamaters with the child's canonical hash so the overall scope hash is stable across sessions
       if (c.type === "SubCircuit" && Array.isArray(comp.properties?.constructorParamaters)) {
-        const params = (comp.properties.constructorParamaters as unknown[]).slice();
+        const params = comp.properties.constructorParamaters.slice();
         const childId = Number(params[0]);
         params[0] = (!isNaN(childId) && childHashes?.get(childId)) ?? params[0];
         comp = { ...comp, properties: { ...comp.properties, constructorParamaters: params } };
--- a/v1/src/simulator/src/data/canonical.ts
+++ b/v1/src/simulator/src/data/canonical.ts
@@ -1,0 +1,1060 @@
+import { circuitElementList } from "../metadata";
+
+type WireNode = {
+  bitWidth: number;
+  connections: WireNode[];
+  deleted?: boolean;
+  type: number;
+  x: number;
+  y: number;
+  [key: string]: unknown;
+};
+
+type NodeIndexMap = Map<WireNode, number>;
+
+type ComponentSaveData = {
+  nodes: Record<string, unknown>;
+  values?: Record<string, unknown>;
+  constructorParamaters?: unknown[];
+};
+
+type CVComponent = {
+  objectType: string;
+  label?: string;
+  bitWidth: number;
+  customSave: () => ComponentSaveData;
+  direction?: unknown;
+  propagationDelay?: number;
+  state?: unknown;
+  labelDirection?: unknown;
+  x?: number;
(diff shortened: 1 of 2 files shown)


v1/src/simulator/src/data/canonical.ts
```diff
@@ -0,0 +1,1060 @@
...
+          .map((c) => Number((c.properties.constructorParamaters as unknown[])?.[0]))
+          .filter((id) => !isNaN(id)),
+      ),
+    ];
+
+    for (const targetId of subcircuitRefs) {
+      if (inDegreeMap.has(targetId)) {
+        dependents.get(targetId)!.push(circuitId);
+      }
+    }
```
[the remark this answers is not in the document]
Harkeerat24: Implementing a fix for this!

v1/src/simulator/src/data/canonical.ts
```diff
@@ -0,0 +1,1067 @@
...
+  while (queue.length > head) {
+    const dequeuedScope = queue[head++]!;
+    topologicalOrder.push(dequeuedScope);
+
+    for (const dep of dependents.get(dequeuedScope) ?? []) {
+      const newIndegree = indegreeMap.get(dep)! - 1;
+      indegreeMap.set(dep, newIndegree);
+      if (newIndegree === 0) {
+        queue.push(dep);
+        queue.sort((a, b) => a - b);
```
aryanndwi123: The array uses a `head` pointer to simulate dequeue, elements at indices `0..head-1` are already consumed, sorting the full array after pushing a new element can move that element to an index below `head` when that happens, `queue[head]` points to an already-processed scope, which gets processed a second time. In the worst case this also causes infinite looping.

The root issue: sorting a slice that includes already-consumed entries is not a safe operation with a pointer-based queue.

v1/src/simulator/src/data/canonical.ts
```diff
@@ -0,0 +1,1067 @@
...
+function wlFingerprint(components: ComponentDraft[]): Map<number, string> {
+  if (components.length === 0) return new Map<number, string>();
+
+  const initialSignatures = new Map<number, string>();
+  for (let i = 0; i < components.length; i++) {
+    const comp = components[i];
+    let signature = `${comp.type}|${comp.bitWidth ?? 1}|${portKey(comp)}`;
+    if (comp.type === "SubCircuit") {
+      const subId = (comp.properties?.constructorParamaters as unknown[])?.[0] ?? "";
+      signature += `|${subId}`;
```
aryanndwi123: `subId` here is the child scope's numeric ID, which is assigned at creation time and is not stable across browser sessions. Two users who build the same subcircuit layout independently will end up with different scope IDs, different WL signatures, and therefore different canonical hashes for structurally identical circuits defeating the core purpose of this pipeline.

What should be used instead is something that is itself stable and content-derived.

v1/src/simulator/src/data/canonical.ts
```diff
@@ -0,0 +1,1067 @@
...
+
+    let indegree = 0;
+    for (const targetId of subcircuitRefs) {
+      if (inDegreeMap.has(targetId)) {
+        indegree++;
+        dependents.get(targetId)!.push(circuitId);
+      }
+    }
+
+    inDegreeMap.set(circuitId, indegree);
```
aryanndwi123: The first loop initialises every valid scope in `inDegreeMap` with indegree 0. The second loop overwrites that value with the real computed indegree. If a scope passes the first loop's guard (`scope.id !== undefined`) but fails the second loop's guard (`!scope.allNodes`), it stays at indegree 0 in `inDegreeMap` and gets placed in `topologicalOrder` as a zero-indegree node, but `pairs.get(id)` returns `undefined`, so it is silently missing from the final `circuits` output with no error.

aryanndwi123 (commented):
Good direction  subcircuit identity in the WL graph and topological ordering are both the right ideas but there are two bugs that must be fixed before merge, and one design decision that will break cross-session hash determinism

v1/src/simulator/src/data/canonical.ts
Harkeerat24: Changed to sort only the unprocessed tail, so consumed indices are never moved.

v1/src/simulator/src/data/canonical.ts
Harkeerat24: Replaced the numeric scope ID with the child's canonical hash in both the WL signature and the hash input. The dependency graph is now built from raw scope data first, then scopes are processed in topological (leaves-first) order so the child's hash is always available before the parent needs it.

v1/src/simulator/src/data/canonical.ts
Harkeerat24: Added !scope.allNodes to the initial collection guard so scopes that won't be processed never enter the dependency graph.

v1/src/simulator/src/data/canonical.ts
```diff
@@ -0,0 +1,1082 @@
...
+    isMainCircuit?: boolean;
+    code?: string;
+    subCircuitScopeIds?: string[];
+  };
+  restrictedCircuitElementsUsed?: string[];
+  [key: string]: unknown;
+};
+
+type ComponentDraft = {
+  id?: string;
```
tachyons, line 61: Why id field is nullable ?

v1/src/simulator/src/data/canonical.ts
```diff
@@ -0,0 +1,1082 @@
...
+
+  const data = new TextEncoder().encode(text);
+  const hashBuffer = await crypto.subtle.digest("SHA-256", data);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  return hashArray.map((byte) => byte.toString(16).padStart(2, "0")).join("");
+}
+
+// Put the below line in the console to get the json
+// await (await import('/simulatorvue/v1/v1/src/simulator/src/data/canonical.ts')).canonicaliseScope(globalScope)
+export async function canonicaliseScope(
```
tachyons, line 885: What is the entry point for this function, do we trigger this for each update in the scope, or when saving ?

v1/src/simulator/src/data/canonical.ts
```diff
@@ -0,0 +1,1082 @@
+import { circuitElementList } from "../metadata";
+
+type WireNode = {
+  bitWidth: number;
+  connections: WireNode[];
+  deleted?: boolean;
+  type: number;
+  x: number;
+  y: number;
+  [key: string]: unknown;
```
tachyons, line 10: Can we remove unknown type as much as possible and make code type safe ?

v1/src/simulator/src/data/canonical.ts
```diff
@@ -0,0 +1,1082 @@
...
+  const childHashes = new Map<number, string>();
+  for (const id of topologicalOrder) {
+    const scope = scopeById.get(id);
+    if (!scope) continue;
+
+    const circuit = await canonicaliseScope(scope, childHashes);
+    pairs.set(id, circuit);
+    childHashes.set(id, circuit.canonicalHash);
+    circuitHashes.push(circuit.canonicalHash);
+  }
```
tachyons, line 1061: Can we memoise canonical hash for subcircuits when it is not changed

tachyons:
fyi @gr455 @satu0king

v1/src/simulator/src/data/canonical.ts
Harkeerat24: ComponentDraft.id is optional because it's used in two stages 

- buildComponentDrafts() creates the draft without an id.
- assignComponentIds() fills in comp.id afterwards, once components have been canonically sorted.

This is important as the id assignment needs to be in canonically sorted order for determinism and better readability.

v1/src/simulator/src/data/canonical.ts
Harkeerat24: The export exists so that test files can import and test the canonicalisation of individual scopes in isolation without going through the full project pipeline. This will also be used in importCanonical.ts for verification of round trip

v1/src/simulator/src/data/canonical.ts
Harkeerat24: Yeah, this is a good feature to add. I will create a new PR specifically for this feature after having a discussion about how to proceed  (The approach I think of is to have a cache and keep track of changes in the circuit)

JoshVarga (approved):

v1/src/simulator/src/data/canonical.ts
Harkeerat24: I have changed and replaced some unknowns with Direction literals and tightened ComponentSaveData types. I don't think other unknown types are required. If you have any changes in mind, please tell me.

Harkeerat24:
@tachyons PTAL

Harkeerat24:
FINALLLYYYY!!!!!!!!!

PR #1112, opened 2026-06-26
feat: add VFS guard to validate Yosys output before parsing

Me-Priyank:
Fixes #1085 

### Describe the changes you have made in this PR -

Added a VFS guard to validate the output from Yosys before trying to parse it. Right now the synthesis worker just blindly calls JSON.parse on whatever Yosys returns. If Yosys crashes silently or returns bad data, it causes confusing errors in the UI. 

This PR adds a small helper function `parseYosysOutput` that checks if `output.json` exists, handles Uint8Array decoding, checks for empty files, and safely wraps JSON.parse. I also added full unit tests for all the validation cases.

**Note:** This branch should only be merged after #1111 and #1105 are merged. The correct merging flow should be #1111 -> #1105 -> then this PR. The diff currently shows 14 files changed with large no. of lines of code, but this will automatically be lowered to just 3 files after these 2 PRs are merged.

### Screenshots of test -

Screenshot 2026-06-26 150737

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [x] No, I wrote all the code myself
- [ ] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [ ] I have reviewed every single line of the AI-generated code
- [ ] I can explain the purpose and logic of each function/component I added
- [ ] I have tested edge cases and understand how the code handles them
- [ ] I have modified the AI output to follow this project's coding standards and conventions
### Explain your implementation approach:

I kept the logic out of the main worker file and put it in a separate pure function `vfsGuard.js`. This makes it much easier to test all the edge cases without needing to mock the whole Yosys WASM engine. The worker just imports this function and passes the raw result object to it. If anything is wrong, the guard throws a clear error message that the worker can catch and send back to the UI.

---

## Checklist before requesting a review
- [x] I have added proper PR title and linked to the issue
- [x] I have performed a self-review of my code
- [x] **I can explain the purpose of every function, class, and logic block I added**
- [x] I understand why my changes work and have tested them thoroughly
- [x] I have considered potential edge cases and how my code handles them
- [x] If it is a core feature, I have added thorough tests
- [x] My code follows the project's style guidelines and conventions

---

Note: Please check **Allow edits from maintainers** if you would like us to assist in the PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synthesis reliability by preventing overlapping synthesis requests and handling worker errors and timeouts more cleanly.
[...]

PR #1116, opened 2026-07-01
feat: add human-readable syntax error messages for Verilog synthesis

### LINKED ISSUE #1113: Improve error handling in client-side synthesis to show human-readable syntax errors
### Description

**Is your feature request related to a problem? Please describe.**

Yes. Currently, when a user enters invalid Verilog code (like a missing `endmodule` or a syntax error), the client-side synthesis worker catches the crash and displays a generic **"Exited with status 1"** error in the UI, along with the raw yosys error which is not human readable.

Image

This happens because the worker is set to silently discard Yosys standard error output (`printErr: silentPrint`). Since we throw away the error logs, users have no idea what actually went wrong with their code, making debugging very frustrating.

**Describe the solution you'd like.**

Instead of silencing the errors, we should capture the standard error output from Yosys. 

When Yosys crashes due to invalid Verilog, it outputs very clear error messages, such as:
`input.v:1 ERROR: syntax error, unexpected end of file`

The solution would be to:
1. Create an array to collect strings from `printErr` while Yosys runs.
2. If Yosys throws an exception (like status 1), read the collected error strings.
3. Extract the actual syntax error message and send that back to the UI instead of the generic exit status.

This will give users immediate, readable feedback on exactly which line caused the problem.

**Describe alternatives you've considered.**

We could try to validate the Verilog code before sending it to Yosys, but that would require building parser in JavaScript. It is much more reliable to let Yosys do the parsing and just accurately report the errors it generates.

**Is requested feature related to vue simulator?**

Yes. This directly affects the user experience in the Vue simulator when using the offline/client-side Verilog synthesis feature.

**Additional context**

This is part of the GSoC'26 Project 7 work for improving the offline synthesis experience in the Tauri desktop app.

Me-Priyank:
Fixes #1113

### Describe the changes you have made in this PR -
Currently, when a user makes a syntax error in their Verilog code, the synthesis engine just crashes and displays a generic "Exited with status 1" message. This happens because the raw error logs from Yosys are dumped straight to the browser console by the WASI shim, bypassing our error handlers.

This PR fixes that by:
1. Intercepting `console.log` and `console.error` during the Yosys synthesis run to capture the raw stderr output.
2. Adding a new `errorParser.js` utility that scans these captured logs, finds the specific line where the error occurred, and translates raw Yosys jargon (like `TOK_ID` or `TOK_ASSIGN`) into plain, easy to understand English.
3. Adding a full suite of unit tests to verify the parser works across different types of syntax errors.

Now, instead of a silent crash, users will see clear messages like "Syntax error on line 3: unexpected end of file (missing 'endmodule' or closing statement?)".

### Screenrecording -

https://github.com/user-attachments/assets/c87daa6d-1d69-4b94-948e-77d2bd6cf8d5

---

### Explain your implementation approach:
The main problem is that WebAssembly component models (specifically the WIT shim used by yowasp) route standard error directly to the browser console. So our worker was never actually seeing the syntax errors.

To fix this, I chose to temporarily override `console.log` and `console.error` right before calling `runYosys`. This lets us push the log lines into a temporary array. Once Yosys finishes or throws an error, I restore the original console functions.

For the parser itself, I created a pure function called `extractYosysError`. It scans the captured log array backwards to find the exact line where the error triggered. Then it maps Yosys internal token names to readable strings using a dictionary. I separated this parser into its own file so I could easily write unit tests for it without needing to mock the entire Web Worker environment.

---

---

## Summary by CodeRabbit

* **New Features**
  * Improved synthesis error messages with clearer, more human-readable wording.

* **Bug Fixes**
  * Syntax errors now include better line-based guidance, including “unexpected end of file” cases.
  * When multiple errors are present, the most relevant/latest error is shown.
  * If no detailed line info is available, a fallback message is displayed instead of a vague failure.

### DIFF
--- a/v1/src/simulator/src/synthesis/clientSynthesis.js
+++ b/v1/src/simulator/src/synthesis/clientSynthesis.js
@@ -2,6 +2,7 @@
 
 var worker = null
 var requestId = 0
+var busy = false
 
 var DEFAULT_SYNTHESIS_TIMEOUT_MS = 30000
 
@@ -10,6 +11,11 @@ export function synthesizeVerilog(verilogCode, onProgress, options = {}) {
     var timeoutMs = options.timeoutMs || DEFAULT_SYNTHESIS_TIMEOUT_MS
 
     return new Promise((resolve, reject) => {
+        if (busy) {
+            reject(new Error('A Verilog synthesis is already in progress'))
+            return
+        }
+
         if (!worker) {
             try {
                 worker = new Worker(new URL('./synthesisWorker.js', import.meta.url), { type: 'module' })
@@ -22,6 +28,7 @@ export function synthesizeVerilog(verilogCode, onProgress, options = {}) {
         var id = ++requestId
         var timer = null
         var currentWorker = worker
+        busy = true
 
         function handleMessage(e) {
             if (e.data.id !== id) return
@@ -51,6 +58,7 @@ export function synthesizeVerilog(verilogCode, onProgress, options = {}) {
         }
 
         function cleanup() {
+            busy = false
             clearTimeout(timer)
             currentWorker.removeEventListener('message', handleMessage)
             currentWorker.removeEventListener('error', handleError)
--- a/v1/src/simulator/src/synthesis/clientSynthesis.js
+++ b/v1/src/simulator/src/synthesis/clientSynthesis.js
@@ -8,7 +8,7 @@ var DEFAULT_SYNTHESIS_TIMEOUT_MS = 30000
 
 // abort + reset the worker if synthesis runs longer (default 30s).
 export function synthesizeVerilog(verilogCode, onProgress, options = {}) {
-    var timeoutMs = options.timeoutMs || DEFAULT_SYNTHESIS_TIMEOUT_MS
+    var timeoutMs = options.timeoutMs ?? DEFAULT_SYNTHESIS_TIMEOUT_MS
 
     return new Promise((resolve, reject) => {
         if (busy) {
--- a/v1/src/simulator/src/synthesis/clientSynthesis.js
+++ b/v1/src/simulator/src/synthesis/clientSynthesis.js
@@ -3,7 +3,12 @@
 var worker = null
 var requestId = 0
 
-export function synthesizeVerilog(verilogCode, onProgress) {
+var DEFAULT_SYNTHESIS_TIMEOUT_MS = 30000
+
+// abort + reset the worker if synthesis runs longer (default 30s).
+export function synthesizeVerilog(verilogCode, onProgress, options = {}) {
+    var timeoutMs = options.timeoutMs || DEFAULT_SYNTHESIS_TIMEOUT_MS
+
     return new Promise((resolve, reject) => {
         if (!worker) {
             try {
@@ -15,6 +20,8 @@ export function synthesizeVerilog(verilogCode, onProgress) {
         }
 
         var id = ++requestId
+        var timer = null
+        var currentWorker = worker
 
         function handleMessage(e) {
             if (e.data.id !== id) return
@@ -38,21 +45,33 @@ export function synthesizeVerilog(verilogCode, onProgress) {
 
         function handleError(err) {
             cleanup()
-            worker.terminate()
-            worker = null
+            currentWorker.terminate()
+            if (worker === currentWorker) worker = null
             reject(new Error('Synthesis worker error: ' + (err.message || 'Unknown error')))
         }
 
         function cleanup() {
-            worker.removeEventListener('message', handleMessage)
-            worker.removeEventListener('error', handleError)
+            clearTimeout(timer)
+            currentWorker.removeEventListener('message', handleMessage)
+            currentWorker.removeEventListener('error', handleError)
         }
 
-        worker.addEventListener('message', handleMessage)
-        worker.addEventListener('error', handleError)
+        currentWorker.addEventListener('message', handleMessage)
+        currentWorker.addEventListener('error', handleError)
+
+        timer = setTimeout(() => {
+            cleanup()
+            currentWorker.terminate()
+            if (worker === currentWorker) worker = null
+            var error = new Error(
+                'Verilog synthesis timed out after ' + (timeoutMs / 1000) + ' seconds'
+            )
+            error.name = 'SynthesisTimeoutError'
+            reject(error)
+        }, timeoutMs)
 
         try {
-            worker.postMessage({ type: 'synthesize', code: verilogCode, id: id })
+            currentWorker.postMessage({ type: 'synthesize', code: verilogCode, id: id })
         } catch (err) {
             cleanup()
             reject(new Error('Failed to send code to synthesis worker: ' + err.message))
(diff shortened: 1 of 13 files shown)


v1/src/simulator/src/synthesis/clientSynthesis.js
```diff
@@ -2,9 +2,20 @@
 
 var worker = null
 var requestId = 0
+var busy = false
+
+var DEFAULT_SYNTHESIS_TIMEOUT_MS = 30000
+
+// abort + reset the worker if synthesis runs longer (default 30s).
+export function synthesizeVerilog(verilogCode, onProgress, options = {}) {
+    var timeoutMs = options.timeoutMs || DEFAULT_SYNTHESIS_TIMEOUT_MS
```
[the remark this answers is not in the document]
Me-Priyank: @coderabbitai This line is not modified in this PR. The timeoutMs || DEFAULT logic is from the timeout/VFS PR (#1105). Also, setting timeoutMs: 0 has no practical use case since a zero-millisecond timeout would always reject before synthesis starts. Will note this for the next PR if needed.

v1/src/simulator/src/synthesis/errorParser.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    return 'Syntax error on line ' + line + ': ' + readable
+}
+
+export function extractYosysError(stderrLines) {
+    if (!stderrLines || !stderrLines.length) return null
+
+    for (var i = stderrLines.length - 1; i >= 0; i--) {
+        var line = stderrLines[i]
```
ThatDeparted2061, line 77: This assumes every captured line is a string, but `line.match` blows up the second one is not. And here is the mean part: this runs inside the worker's `catch`, so if it throws, no error message gets posted at all and the UI just sits there until the 30s timeout. That is the exact failure path this whole feature exists to serve. A cheap `if (typeof line !== 'string') continue` at the top of the loop keeps it honest.

v1/src/simulator/src/synthesis/errorParser.js
```diff
@@ -0,0 +1,88 @@
...
+        var match = line.match(/^input\.v:(\d+):\s*ERROR:\s*(.+)/)
```
ThatDeparted2061: That `^` anchor is doing a lot of trust falls. It only matches when the log line starts with exactly `input.v:`. If the WASI shim adds any leading padding or prefix (console routing loves doing that), this misses, you drop through to the generic `ERROR:` branch, and you lose both the line number and the friendly "missing endmodule" hint. Dropping the `^` (or trimming the line first) lets it match `input.v:3: ERROR:` anywhere in the string. Worth eyeballing one real Yosys stderr dump to confirm the exact shape.

v1/src/simulator/src/synthesis/synthesisWorker.js
```diff
@@ -18,26 +22,35 @@ self.onmessage = async function (e) {
...
+        console.error = function (msg) { stderrLines.push(String(msg)) }
+
+        try {
+            var result = await runYosys(
+                [
+                    '-p',
+                    'read_verilog input.v; setattr -mod -unset top; hierarchy -auto-top; proc; opt; memory -nomap; wreduce -memx; opt -full; write_json output.json',
+                ],
+                { 'input.v': code },
+                { print: silentPrint, printErr: function (msg) { stderrLines.push(msg) } }
```
ThatDeparted2061: Small inconsistency with real teeth: the two `console` overrides right above push `String(msg)`, but this one pushes `msg` raw. If yowasp ever feeds `printErr` a non-string, it lands in `stderrLines` unstringified and later trips the parser's `.match` (see the comment over in errorParser.js). Wrap it in `String()` like its siblings so everything stays a string.

v1/src/simulator/src/synthesis/synthesisWorker.js
```diff
@@ -18,26 +22,35 @@ self.onmessage = async function (e) {
...
-        )
-
-        var jsonContent = result['output.json']
-        if (jsonContent instanceof Uint8Array) {
-            jsonContent = new TextDecoder().decode(jsonContent)
+
+        // Intercept console output; WASI shim routes stderr through console.log
+        var origLog = console.log
+        var origError = console.error
+        console.log = function (msg) { stderrLines.push(String(msg)) }
```
ThatDeparted2061: Heads up: these only grab the first argument. If the shim ever calls `console.log('input.v:3:', 'ERROR: ...')` in pieces, everything after the first vanishes and the parser never sees the ERROR. `function (...args) { stderrLines.push(args.map(String).join(' ')) }` future-proofs it for basically nothing.

v1/src/simulator/src/synthesis/vfsGuard.js
```diff
@@ -0,0 +1,31 @@
...
+    if (typeof raw !== 'string') {
+        throw new Error('Yosys output.json has an unsupported type: ' + typeof raw)
+    }
+
+    if (!raw.trim()) {
+        throw new Error('Yosys produced an empty output.json.')
+    }
+
+    try {
+        return JSON.parse(raw)
```
ThatDeparted2061: Flagging this again since it rides along from #1112 and should really be fixed back there: this promises to return a netlist, but `JSON.parse` happily hands back non-objects. Give `output.json` a value of `'null'`, `'42'`, or `'"oops"'` (all valid JSON) and this returns `null` / `42` / `'oops'` without a peep, then `yosys2digitaljs` chokes on it and throws something unreadable, which is the exact cryptic failure this guard is meant to stop. One check closes it:

```js
var parsed
try {
    parsed = JSON.parse(raw)
} catch (err) {
    throw new Error('Yosys produced invalid JSON in output.json: ' + err.message)
}
if (parsed === null || typeof parsed !== 'object') {
    throw new Error('Yosys output.json was valid JSON but not a netlist object.')
}
return parsed
```

Since this file is shared, best to land the fix in #1112 so it flows through here rather than patching it twice.

v1/src/simulator/src/synthesis/errorParser.js
Me-Priyank: fixed

v1/src/simulator/src/synthesis/synthesisWorker.js
Me-Priyank: fixed

v1/src/simulator/src/synthesis/synthesisWorker.js
Me-Priyank: fixed

v1/src/simulator/src/synthesis/vfsGuard.js
Me-Priyank: fixed

v1/src/simulator/src/synthesis/errorParser.js
Me-Priyank: fixed

Me-Priyank:
@ThatDeparted2061 @092vk @Nihal4777 PTAL

ThatDeparted2061:
Conflicts Priyank

Me-Priyank:
Solving .. jus give me 2 minutes

Me-Priyank:
@ThatDeparted2061 resolved

Issue #469, opened 2025-02-10
Chore: Reduce usage of set timout and set interval


Arnabdaz:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Describe the bug

In the vue js code base there are places where we used set interval or set timeout, we want to replace them with [vue lifecycle hooks](https://vuejs.org/guide/essentials/lifecycle) which would reduce the load on browsers for end users.

### Steps to Reproduce

N/A

### Expected Behavior

The usage of set timeout and set interval should be minimized in the vue files.

### Screenshots

N/A

### Is the faced issue/bug related to the Vue simulator?

Yes

### Used Vue simulator with or without backend?

None

### Is the bug present only on the dev server, the build, or both?

Both

### Device Information

```markdown
N/A
```

### Are you working on this issue?

No

rxiddhi:
Please assign me this issue, I can work on it.

Danish0111:
I want to work on it

Arnabdaz:
This issue won't be assigned to anyone its a open issue, just put the name of the file you are working on here so that others can know.

rxiddhi:
I am working on the following files-
ContextMenu.vue
CombinationalAnalysis.vue
ExportProject.vue 
ApplyThemes.vue

Arnabdaz:
Only take one file at time @rxiddhi so that others also have a chance to work as its a good first issue.

rxiddhi:
sure @Arnabdaz , i will be working on CombinationalAnalysis.vue

ThatDeparted2061:
done, ApplyThemes.vue in #441

Rustix69:
@Arnabdaz Done with the Properties Panel.

rxiddhi:
@Arnabdaz Hi, I did the required changes, but they got merged with my previous PR. I want to create a separate PR for them. Could you guide me on the best way to do this? #465

Jay2006sawant:
Hello  @Arnabdaz Would be working on ExportProject.vue

blazethunderstorm:
@Arnabdaz i would be working on contextMenu.vue file

blazethunderstorm:
@Arnabdaz  thanks would make a pr for it

Aakansha-Chavan:
[...]

PR #1127, opened 2026-07-06
refactor: clear setInterval on unmount in PropertiesPanelMobile.vue

Shruti2110-coder:
## Description
`PropertiesPanelMobile.vue` was using `setInterval(showPropertiesPanel, 100)` inside `onMounted` without ever clearing it. This causes the interval to keep running indefinitely even after the component unmounts, and creates a new overlapping interval each time it remounts — a memory/performance leak.

## Changes
- Store the interval ID in a variable
- Clear it in `onUnmounted`

## Testing
Tested locally on mobile viewport — properties panel still updates correctly when selecting elements, layout mode, and project properties. No regressions observed.

Relates to #469

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mobile properties panel behavior by stopping a background update timer when the panel is closed or destroyed.
  * Prevents unnecessary activity after navigating away, helping avoid lingering updates and improving app stability.

### DIFF
--- a/src/components/Panels/PropertiesPanel/PropertiesPanelMobile.vue
+++ b/src/components/Panels/PropertiesPanel/PropertiesPanelMobile.vue
@@ -141,7 +141,7 @@
 </template>
 
 <script lang="ts" setup>
-import { toRaw, onMounted, computed } from 'vue'
+import { toRaw, onMounted, onUnmounted, computed } from 'vue'
 import { useSimulatorMobileStore } from '#/store/simulatorMobileStore'
 import ProjectProperty from './ModuleProperty/ProjectProperty/ProjectProperty.vue'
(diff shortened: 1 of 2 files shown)


salmoneatenbybear (approved):
Hey this change looks fine to me, my understanding is that before this PR the component started a timer every time it mounted, but never stopped it. After this PR, vue stops the timer when component is destroyed. 

One thing to flag is that desktop still has the old pattern in `src/components/Panels/PropertiesPanel/PropertiesPanel.vue:21`. 

That is not a blocker for this PR because PR 1127 specifically claims PropertiesPanelMobile.vue, but someone should fix desktop separately. Maybe u can raise a PR for that too, if my flag is valid. 

CC : @CircuitVerse/simulator-reviewers

PR #1130, opened 2026-07-06
fix: remove duplicate nested v-for and improve Testbench Creator layout

### LINKED ISSUE #1128: 🐞 Bug: Testbench results column renders duplicate ticks/crosses due to nested v-for loop, overflowing dialog

### Describe the bug

In the Testbench Creator dialog, the "Results" column (shown after running a test) renders far more pass/fail marks (✔/✘) per row than there are actual outputs. For a circuit with N outputs, it renders N × N marks instead of N. This causes the results column to visibly overflow past the dialog's right edge instead of being contained.

Root cause: `TestBenchCreator.vue` has two nested `v-for` loops iterating over the exact same `results[groupIndex]` array an outer loop wrapping an inner loop that does the same iteration again, so the real per-output check gets duplicated once for every output instead of rendering once per output.

### Concrete example
With a circuit that has **4 outputs** (out1-out4):
- The outer loop runs 4 times (once per output).
- For **each** of those 4 outer passes, the inner loop runs 4 times again.
- Total marks rendered per row = 4 × 4 = **16**, instead of the correct **4** (one per output).

This is the same reason a nested loop like:
​```js
for (let i = 0; i < 4; i++) {
  for (let j = 0; j < 4; j++) {
    render(mark);
  }
}
​```
renders "mark" 16 times, not 8 or 4 — the inner loop fully repeats on every single outer iteration rather than running once total. This is what causes the visible pile of extra ✔/✘ marks overflowing past the dialog's edge.

### Steps to Reproduce

1. Build any circuit with a few Input/Output elements (all labeled correctly, e.g. `inp1`-`inp4`, `out1`-`out4`).
2. Open Testbench, create a test with 4 inputs and 4 outputs, add a few test case rows, and attach it.
3. Click "Edit" (or reopen the Testbench Creator) with results showing.
4. Observe the "Results" column for each row — count the ticks/crosses shown per row (should be 16, not 4)<--Bug.

### Expected Behavior

Each row's Results column should show exactly one tick/cross per output (4 outputs = 4 marks per row), matching `out1` through `out4` one-to-one. It should not repeat the full set of checks once per output.

### Screenshots

https://github.com/user-attachments/assets/33e0c46a-c619-403d-84de-349cbc7f2a0a

There should be only 4 crosses but its rendering 16 (Which are not visible to us but they are 16 crosses) due to duplicate loop. And when this happens the Cross tend to overflow which is a major UI bug.

Yes

### Vue simulator version

v1

Without Backend

Both

### Device Information

```markdown
OS: Windows 11
Browser: Chrome
Version: v1
```

### Additional Context

_No response_

Yes

DevPatils:
Fixes #1128

- Removed a duplicated nested `v-for` loop in the Testbench Creator's results column (`TestBenchCreator.vue`, both v0 and v1) that caused N² result marks to render per row instead of N  e.g. 4 outputs produced 16 marks instead of 4, overflowing the dialog. This was the original bug in #1128.
- While testing the fix, found and resolved several related layout issues in the same component that only showed up with varying numbers of inputs/outputs:
  - Input/Output/Bitwidth columns previously used fixed-width CSS grid columns that broke (numbers disappearing, text truncating, columns misaligned between the header/Label/Bitwidth rows and the actual data rows) once you had more than a couple of inputs or outputs.
  - Replaced this with dynamically computed column widths based on actual item count (with a 5-item baseline so the layout doesn't look cramped with very few items), applied consistently across the header, Label, Bitwidth, and every data row via a shared computed style, so all rows stay column-aligned.
  - Each Input, Output, and Results box now scrolls independently within its own bounds when it has more items than fit, instead of the columns squeezing content into invisibility or the whole dialog overflowing.
  - Added per-Group "Input"/"Output"/"Results" column headers for clarity.
  - Removed the "Results" pill from the top header/Label/Bitwidth section, since that section never actually displays result data — Input/Output now use that freed space instead.

---

The original bug was a copy-paste artifact: the results column had two nested loops iterating over the same array, so instead of one pass-or-fail mark per output, it rendered outputs² marks per row. The fix was simply removing the duplicate outer loop and keeping the single correct one.

While verifying the fix by actually building test circuits with different input/output counts (1, 4, 7+), I found the surrounding layout didn't handle varying counts well at all — columns used fixed-width CSS grid tracks, so numbers and controls would get squeezed to invisibility or misalign between rows once you had more than a few inputs/outputs. I went through several iterations to get this right: first tried letting columns grow freely with content, which caused the whole dialog to need horizontal scrolling; then tried a single unified scrollbar for the whole table, which felt too broad; landed on computing each column's width from the actual item count (with a 5-item minimum baseline) and giving each individual box (Input, Output, Results) its own independent scrollbar via CSS Grid's `minmax(0, ...)` track sizing plus `overflow-x: auto`, so scrolling through many inputs doesn't affect the Output or Results box. Results also got its own smaller per-cell width since a checkmark needs far less room than a labeled input box, so it rarely needs to scroll at all in practice.

I considered making the whole table scroll as a single unit instead of per-box, but that ties Input/Output/Results scroll positions together in a way that felt worse for actually reading the data side by side, so I kept them independent.

---

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the creator dialog’s grid column sizing across Inputs, Outputs, Results, and Bitwidth for more consistent alignment.
  * Made results markers (check/cross) more reliable when some result data is missing or incomplete.
  * Refreshed horizontal overflow and fixed-width cell behavior so wider tables stay usable and better aligned.

### DIFF
--- a/src/components/Panels/TestBenchPanel/TestBenchCreator.vue
+++ b/src/components/Panels/TestBenchPanel/TestBenchCreator.vue
@@ -33,7 +33,7 @@
                 </div>
 
                 <div v-else class="data-table-container">
-                    <div class="data-grid header-grid" :class="{ 'with-results': testBenchStore.showResults }">
+                    <div class="data-grid header-grid" :style="{ gridTemplateColumns: topGridColumnsStyle }">
                         <div class="grid-cell header-cell label-col"></div>
                         <div class="grid-cell header-cell inputs-col">
                             <span>Inputs</span>
@@ -43,13 +43,9 @@
                             <span>Outputs</span>
                              <v-btn icon size="x-small" variant="flat" class="ml-2 action-icon" @click="increOutputs"><v-icon>mdi-plus</v-icon></v-btn>
                         </div>
-                        <div v-if="testBenchStore.showResults" class="grid-cell header-cell results-col">
-                            Results
-                        </div>
                     </div>
-                </div>
 
-                    <div class="data-grid labels-grid" :class="{ 'with-results': testBenchStore.showResults }">
+                    <div class="data-grid labels-grid" :style="{ gridTemplateColumns: topGridColumnsStyle }">
                         <div class="grid-cell label-col">
                             Label
                         </div>
@@ -67,7 +63,7 @@
                         </div>
                     </div>
 
-                    <div class="data-grid bitwidth-grid" :class="{ 'with-results': testBenchStore.showResults }">
+                    <div class="data-grid bitwidth-grid" :style="{ gridTemplateColumns: topGridColumnsStyle }">
                       <div class="grid-cell label-col">Bitwidth</div>
                       <div class="grid-cell inputs-col">
                         <div v-for="(bw, i) in inputsBandWidth" :key="`in-bw-${i}`" class="io-cell bitwidth-row">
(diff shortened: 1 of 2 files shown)


DevPatils:
Before : The looping caused the ticks/crosses overflow

https://github.com/user-attachments/assets/016d7213-0651-490b-b398-49721fc0aa10

After: Fixed the loop so it demonstrates correct output symbols(tick/cross) and made the UI look consistent. 

https://github.com/user-attachments/assets/dcfed1a9-cf6a-4f1f-ad9e-9f150f289410

DevPatils:
@Nihal4777 @tachyons @ThatDeparted2061 Please review my PR, I have resolved all the CodeRabbit checks  :)

ThatDeparted2061:
The bug is a nice catch, but come on man, don't use so much AI that even the PR title is not complete, fix that

DevPatils:
> The bug is a nice catch, but come on man, don't use so much AI that even the PR title is not complete, fix that

Ohhh Damnnnn how can I make this mistake, I'll be aware from now onwards  :)
But are the changes correct? @ThatDeparted2061

PR #1131, opened 2026-07-06
feat: add canonical JSON import pipeline with round-trip verification

### LINKED ISSUE #1071: (GSoC'26) Week 4: Write canonicalSchema.json. Integrate validation in importCanonical.js.

Harkeerat24:
Fixes #1071 

Adds a complete canonical JSON import pipeline (`importCanonical.ts`) that can reconstruct a full project from the canonical JSON format produced by `canonicaliseProject()`.

  **Key files:**
  - `v1/src/simulator/src/data/importCanonical.ts` — New file, the entire import pipeline
  - `v1/src/simulator/src/data/canonical.ts` — Minor tweaks to support the import round-trip

NaN

---

- [ ] Yes, I used AI assistance

The import processes circuits in topological order (dependencies first), so subcircuits are always fully constructed before the parent circuit tries to reference them. For each circuit:

  1. `computeImportOrder()` — Builds a dependency graph from SubCircuit references in the canonical JSON and runs Kahn's algorithm to determine import order.

  2. `buildComponents()` — Constructs every component instance (gates, SubCircuits, inputs/outputs, etc.) from the JSON's component definitions. Each component is positioned using its layout data from the canonical format. SubCircuit instances use their `constructorParamaters` to find and reference the correct sub-scope.

  3. `wireComponents()` — Connects component ports together through shared nets as described in the JSON's netlist. Nets with intermediate routing nodes are handled separately.

  4. `restoreDefaultState()` — Restores saved initial values for stateful components (FlipFlops, etc.).

  5. `restoreIntermediateNodes()` — Creates junction nodes and restores wire routing (bends, intermediate points) from the canonical layout data.

  6. `restoreScopeMetadata()` — Copies visual metadata (name, canvas viewport, subcircuit symbol dimensions, Verilog flags) back onto the scope.

  7. `verifyRoundTrip()` — Re-exports the imported scope and compares its canonical hash to the original

  8. `importCanonical()` — The main entry point. Orchestrates the above functions, handles scope re-ID to map the imported project's circuits onto the user's current workspace, and triggers canvas refresh + simulation propagation.
---

- [x] I have added a proper PR title and linked to the issue

## Summary by CodeRabbit

- **Bug Fixes**
  - Tightened canonical project import validation: import now reports success only when there are no build, wire, routing, or verification errors.
  - On import failure, the app reliably attempts to restore the prior circuit snapshot and, on success, displays the import-failure message with all error details combined.
  - Canvas refresh and simulation/engine updates are triggered only after a fully successful import.
- **Localization**
  - Removed the obsolete generic “invalid file” import message key from English, Bengali, and Hindi translations.


Harkeerat24:
@JoshVarga @aryanndwi123 @tachyons PTAL:)

v1/src/simulator/src/data/importCanonical.ts
```diff
@@ -0,0 +1,728 @@
...
+      if (!scopeMap.has(subcircuitId)) {
+        errors.push(`SubCircuit "${id}": scope ${subcircuitId} not found`);
+        continue;
+      }
+
+      const Constructor =
+        (modules as Record<string, ComponentConstructor | undefined>)["SubCircuit"] ||
+        (SubCircuit as unknown as ComponentConstructor);
+      try {
+        instance = new Constructor(pos.x, pos.y, scope, String(subcircuitId));
```
tachyons, line 162: What will the UX flow when a sub circuit failed to initialise ?

tachyons:
- Using `unknown` type should be as an exception, those cases should be documented . Also when concrete types are not feasible, use interfaces for shared properties like  label, propagationDelay, state etc

v1/src/simulator/src/data/importCanonical.ts
```diff
@@ -0,0 +1,728 @@
...
+    }
+
+    // Junction-to-junction connections
+    for (let i = 0; i < edges.length; i++) {
+      const [fromId, toId] = edges[i];
+      const fromNode = junctionNodes[fromId];
+      const toNode = junctionNodes[toId];
+      if (fromNode && toNode) {
+        try {
+          fromNode.connect(toNode);
```
tachyons, line 383: Does this individual call re-render circuit ?

tachyons:
@coderabbitai Could you review this PR for 

- Idiomatic typescript
- Code readability
- Perfomance
- Type saftey

tachyons:
@coderabbitai help

tachyons:
@coderabbitai generate unit tests

v1/src/simulator/src/data/importCanonical.ts
Harkeerat24: No, the `connect()` function calls `updateCanvasSet(true)`, `updateSimulationSet(true)`, `scheduleUpdate()`. They're all boolean flag setters. The render happens at the end of `importCanonical()`

v1/src/simulator/src/data/importCanonical.ts
Harkeerat24: When a subcircuit is failed to intialise then the respective error is passed to ImportProject.vue, which is displayed in red font in the open dialog saying something like "[1002] SubCircuit "SubCircuit_0": scope 1002 not found".

tachyons:
@coderabbitai review this code for idiomatic vue 3 code

tachyons:
@coderabbitai generate unit tests

Harkeerat24:
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> fd "ImportProject.vue" -a
> echo "---"
> cat -n v1/src/components/DialogBox/ImportProject.vue 2>/dev/null | head -300
> ```
> 
> Length of output: 9790
> 
> `@tachyons`, here's a review of `v1/src/components/DialogBox/ImportProject.vue` from a Vue 3 idioms perspective:
> 
> ### 1. Manual DOM manipulation instead of reactive state
> `addDropFile` and every error branch in `receivedText` do:
> 
> ```ts
> document.querySelector('.fileInput')?.classList.add('error--text')
> ```
> 
> This bypasses Vue's reactivity model entirely — it's imperative DOM mutation in a component that otherwise uses `ref`s. A more idiomatic Vue 3 approach is to derive the error class from state:
> 
> ```ts
> const hasError = computed(() => !!errorMessage.value)
> ```
> 
> ```
> <v-file-input
>   :class="{ 'error--text': hasError }"
>   ...
> >
> ```
> 
> This also removes the need to repeat `document.querySelector('.fileInput')?.classList.add(...)` six times across the file — a single `errorMessage.value = t(...)` assignment would be enough to update both the message and the styling.
> 
> ### 2. `document.getElementById` / `querySelector` instead of template refs
> ```ts
> document.getElementById('fileInput')?.click()
> ```
> 
> Prefer a `ref` on the element/component and call the native trigger through it (e.g., `fileInputRef.value?.$el.querySelector('input')?.click()`, or better, check if Vuetify's `v-file-input` exposes a `click`/`browse` method via its instance). Using `id`/class selectors couples the logic to markup/CSS classes rather than the component instance, which is fragile if the template changes.
> 
> ### 3. `(window as any).globalScope`
> ```ts
> const activeScope = (window as any).globalScope
> ```
> 
> This `any` cast defeats type safety, which was explicitly called out in this PR's review thread. If `globalScope` must live on `window` for legacy interop, at least declare it via a `declare global { interface Window { globalScope: Scope } }` or expose it through a small composable/store getter so components consume a typed value instead of casting inline.
> 
> ### 4. Mixed `<script>` + `<script setup>` blocks
> ```ts
> <script lang="ts">
> export function ImportProject() {
>     const SimulatorState = useState()
>     SimulatorState.dialogBox.import_project_dialog = true
> }
> </script>
> ```
> 
> Exporting an imperative "open dialog" function from the component file works, but it's not very idiomatic for Vue 3's Composition API style used elsewhere in this file. Consider moving this into the `SimulatorStore` itself as an action (e.g., `SimulatorState.openImportDialog()`), so callers depend on the store rather than importing a function from a `.vue` file. This keeps the component file focused purely on rendering + local reactive logic.
> 
> ### 5. Ad-hoc `watch`-based polling for file selection
> ```ts
> fileWatchStop = watch(
>     () => file.value,
>     () => { if (getFileInstance()) { ... readFile() } }
> )
> ```
> 
> Watching a ref just to detect "a file was chosen after we programmatically clicked the hidden input" works, but it's a workaround for not having a direct `@update:model-value` / `change` handler on `v-file-input`. Since `v-file-input` already emits update events, binding directly to `@update:model-value="onFileSelected"` (only calling `readFile` when a file is present) would remove the need for a manual `watch`/`fileWatchStop` lifecycle entirely, and it avoids the two-branch `if (!getFileInstance()) { ...watch... } else { readFile() }` split.
> 
> ### 6. `file` ref typing / dual array-or-single handling
> ```ts
> const file = ref(Array<File>())
> ...
> const f = file.value[0] || file.value
> return f instanceof File ? f : null
> ```
> 
> This `f.value[0] || f.value` fallback suggests the type isn't fully known at the call site. Vuetify 3's `v-file-input` v-model is consistently `File[]` (or `File | File[]` depending on `multiple`), so typing `file` precisely (e.g., `ref<File[]>([])`) and always reading `file.value[0]` would remove the duck-typing branch and improve type safety.
> 
> Overall, the biggest wins here would be: (a) replacing manual class/DOM mutation with computed/reactive bindings, (b) using a template ref instead of `getElementById`, and (c) driving the "wait for file selection" flow off the input's own emitted event rather than a manual `watch`.
> 
> ✅ Action performed


@coderabbitai— thanks for the detailed review.

I've implemented 1, 3, and 6. Quick note on the ones I didn't change:

2 (template ref over getElementById): Vuetify 3's v-file-input doesn't expose a stable click() method on its instance for triggering the native picker, and the internal <input> ref path changes across Vuetify minor versions. document.getElementById is the most reliable approach here for now.

 4 (store action over dual script blocks): The standalone ImportProject() export is consumed by external callers outside this component's rendering tree. Extracting it into the store would blur the data-layer boundary.

 5 (@update:model-value over watch): The watch is scoped to a narrow UX flow — clicking "Import" without a file selected opens the picker and auto-imports on selection. Binding @update:model-value would fire on every file change (including clearing), requiring additional guards. The current watch is more explicit about this specific one-shot interaction.

Harkeerat24:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> component="$(fd -a -t f '^ImportProject\.vue$' | head -n 1)"
> printf '%s\n' "=== component: $component ==="
> ast-grep outline "$component" --items all || true
> printf '%s\n' "=== ImportProject.vue ==="
> cat -n "$component"
> 
> printf '%s\n' "=== ImportProject symbol usages ==="
> rg -n -C 3 '\bImportProject\s*\(' -g '*.ts' -g '*.vue' -g '*.js' -g '*.jsx' -g '*.tsx' .
> 
> printf '%s\n' "=== globalScope Window declarations/usages ==="
> rg -n -C 3 'interface\s+Window|globalScope' -g '*.ts' -g '*.d.ts' -g '*.vue' .
> ```
> 
> Length of output: 111958
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> file="v1/src/components/DialogBox/ImportProject.vue"
> 
> printf '%s\n' "=== $file ==="
> cat -n "$file"
> 
> printf '%s\n' "=== PR diff for $file ==="
> git diff --no-ext-diff -U20 HEAD^ HEAD -- "$file" || true
> 
> printf '%s\n' "=== relevant expressions in the PR component ==="
> rg -n -C 4 \
>   'computed\s*\(|error--text|fileWatchStop|watch\s*\(|getElementById|update:model-value|ref<.*File|window\.globalScope|export function ImportProject' \
>   "$file"
> ```
> 
> Length of output: 19177
> 
> `@Harkeerat24` Thanks—verified against `v1/src/components/DialogBox/ImportProject.vue`.
> 
> * **1 and 3:** addressed. Error presentation is now reactive via `hasError`, and `window.globalScope` is properly declared as `ScopeLike | undefined`.
> * **2:** accepted. Given Vuetify’s lack of a stable public picker-trigger API, the ID-based native input trigger is a reasonable compatibility boundary here.
> * **4:** accepted. `ImportProject()` is called through the external listener bridge, so retaining the small exported adapter avoids coupling that integration to the store.
> * **6:** `file` is now correctly `ref<File[]>([])`. Since that makes `file.value` always an array, `getFileInstance()` can eventually be simplified to `return file.value[0] ?? null`; the `val instanceof File` branch is now unreachable.
> 
> For **5**, the one-shot watcher is reasonable for this specific UX flow, and stopping it on selection/retry prevents duplicate imports. One small lifecycle edge remains: if the file picker is dismissed, the watcher stays active until another Import click. Please stop `fileWatchStop` in `onUnmounted` (and optionally when the dialog closes) so it cannot outlive the dialog instance.
> 
> ✏️ Learnings added

@coderabbitai 
I have implemented your suggestion. The watcher now gets cleaned up in onUnmounted as well.
For 6. changing it to `return file.value[0] ?? null` breaks the import button, so I didn't implement this.

Harkeerat24:
cc: @tachyons

PR #1133, opened 2026-07-08
fix: correct node deletion to clear it from its actual parent's nodeList

### LINKED ISSUE #945: 🐞 Bug: Fix Potential Bug in Node Deletion Logic

### Describe the bug

In `src/simulator/src/node.js` at line 940, there's a comment indicating uncertainty about the node deletion logic: `// Hope this works! - Can cause bugs`. This suggests the `nodeList` filtering during node deletion may have edge cases that aren't properly handled.

## Current Behavior

- Node deletion filters `root.nodeList` but the comment indicates uncertainty
- Potential for bugs if the filtering logic doesn't work correctly
- No proper error handling or validation

### Steps to Reproduce

**File:** `src/simulator/src/node.js`

**Line 940:** The problematic code
this.parent.scope.root.nodeList = this.parent.scope.root.nodeList.filter(x => x !== this) // Hope this works! - Can cause bugs

### Expected Behavior

- Node deletion from `root.nodeList` should work reliably
- Proper error handling and validation
- Remove the uncertain comment after verifying the fix works correctly
- Ensure node is properly removed from all relevant lists

### Screenshots

_No response_

Yes

### Vue simulator version

v0

None

None

### Device Information

```markdown

```

### Additional Context

_No response_

Yes

bhuvan-somisetty:
Fixes #945

`Node.delete()` in `src/simulator/src/node.js` had this line, flagged in the issue as uncertain:

```js
this.parent.scope.root.nodeList = this.parent.scope.root.nodeList.filter(x => x !== this) // Hope this works! - Can cause bugs
```

The problem: it always filters `scope.root.nodeList`, no matter what the node's actual parent is. `scope.root` is only the parent for intermediate wire nodes (the ones created directly on the canvas, not owned by a gate/element), and those never get pushed into any `nodeList` in the first place - the constructor only pushes non-intermediate nodes, and only onto `this.parent.nodeList`, not `scope.root.nodeList`.

So for the common case - deleting an input/output node that belongs to an actual circuit element (AndGate, TflipFlop, etc.) - this line was filtering an array the node was never a member of. It silently did nothing, and the deleted node stayed behind in its real parent's `nodeList` forever. Since `nodeList` gets iterated all over `circuitElement.js` (`update()`, `checkHover()`, Verilog generation, bit-width updates, etc.), a stale reference sitting there is exactly the kind of thing that can surface as odd bugs later.

Fix: filter `this.parent.nodeList` directly instead, guarded with a `!== undefined` check the same way the constructor already guards the push. This mirrors the existing (correct) pattern in the `replace()` function further up in the same file. Removed the "hope this works" comment since the behavior is now verified with a test instead of a hope.

Applied the identical fix to `v1/src/simulator/src/node.js` to keep both copies consistent, following the same file it was reported against.

Added `src/simulator/spec/node.spec.js` covering:
- a deleted node is actually removed from its parent element's `nodeList`
- a deleted node is removed from `scope.allNodes` and `scope.nodes`
- deleting a node whose parent has no `nodeList` doesn't throw

Ran the full test suite for both `src` and `v1` projects locally - all passing (128/128).

N/A - this is an internal bookkeeping fix in node deletion, no UI change.

---

The bug is that `Node.delete()` was cleaning up the wrong list. Every `CircuitElement` (gates, flip-flops, etc.) keeps its own `nodeList` of the input/output nodes that belong to it - that list gets populated in the `Node` constructor. But the cleanup on delete assumed every node's parent was `scope.root`, which is only true for intermediate nodes on wires. Those intermediate nodes are explicitly skipped when populating `nodeList` in the first place, so the old cleanup line was a no-op for the case it needed to actually handle. I checked how `replace()` elsewhere in the file does this same kind of cleanup correctly (`parent.nodeList.filter(...)` using the real parent) and matched that pattern, then added the `!== undefined` guard for elements that may not define `nodeList`. Verified with a small test that deletes a node and checks it's actually gone from the parent's `nodeList`, then ran the existing suite to make sure nothing else depends on the old (broken) behavior.

---

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Node deletion now correctly removes nodes from the active parent list without errors when that list is unavailable.
  * Deleting a node also continues to clear it from the simulator’s node collections.

* **Tests**
  * Added coverage for node deletion behavior, including successful removal and handling of missing parent node lists.


bhuvan-somisetty:
Hi @tachyons @ThatDeparted2061 @Nihal4777 @Me-Priyank 

Would appreciate a review on this when you have a moment. Fixes #945 — `Node.delete()` was cleaning up the wrong list (`scope.root.nodeList` instead of the node's actual parent's `nodeList`), so deleted nodes were silently left behind in their owning element's `nodeList` forever. Added tests covering the fix. Thanks!

bhuvan-somisetty:
Hey @092vk, could you take a look at this when you get a chance? Same fix approach as #1129, just applied to the node deletion path this time.

bhuvan-somisetty:
Hi @tachyons @ThatDeparted2061 @Nihal4777, gentle nudge on this one — it's a small node-deletion fix, all checks are green and it's been approved. Whenever you have a moment, it should be good to merge. Thanks!

PR #1207, opened 2026-08-15
#1186 dev container and codespaces setup

### LINKED ISSUE #1186: Add a dev container and Codespaces setup for the Vue simulator
**Overview**

The Vue simulator repo has no dev container, so contributors cannot use Codespaces or "Reopen in Container" and must set up Node by hand. This adds a reproducible one-click setup, mirroring the dev container work in the main CircuitVerse repo.

**Gaps**

 - No dev container: there is no .devcontainer folder at all, so Codespaces and "Reopen in Container" are unsupported.
 - No pinned Node version: no .nvmrc, .node-version, or engines field; only CI declares Node 22, so local setups can drift.
 - Fragile manual setup: Windows users must enable Developer Mode and git symlinks before cloning (for v0 and v1 version switching), which is easy to get wrong.

**Scope**

 - Add a .devcontainer (config plus Dockerfile) on Node 22 to match CI; no database needed.
 - Install dependencies on create and forward the Vite dev port (4000).
 - Add recommended editor extensions and verify symlink-based version switching works inside the container.
 - Add a CI check that validates the dev container, and update the README.

**Acceptance criteria**

 - A fresh Codespace or "Reopen in Container" serves the simulator on port 4000.
 - The test suite and lint run inside the container.
 - Switching between v0 and v1 works in-container.

shivam-G05:
Add a .devcontainer with a Node 22 image matching CI, non-root by default, npm ci on create, and port 4000 forwarded. Bind the Vite dev server to all interfaces only inside the container (DEV_CONTAINER env), since Vite's default localhost-only binding is unreachable through a forwarded port.

Fixes #1186

This repo had no dev container, so Codespaces and "Reopen in Container" were not supported, and contributors had to set up Node by hand with no version pinned anywhere except CI. This PR adds a working dev container.

What changed:

- Added `.devcontainer/Dockerfile`, based on the official Node 22 dev container image, which already runs as a non-root user out of the box.
- Added `.devcontainer/devcontainer.json`, which builds that Dockerfile, sets the remote user explicitly to the non-root user, installs dependencies with `npm ci` on create, and forwards port 4000 with a clear label.
- Added the `Vue.volar` extension recommendation, matching the one already listed in this repo's own `.vscode/extensions.json`.
- Made one small change to `vite.config.ts`. While testing this live, I found that Vite binds to localhost only by default, which means the dev server is not reachable from outside the container even with the port forwarded correctly. I made the server also bind to all interfaces, but only when a `DEV_CONTAINER` environment variable is present, which is only ever set inside the container itself. Every other environment, local development, CI, Netlify, and the desktop build, is unaffected.

While testing, I also found that this repo's `v0/src` folder is meant to be a symlink to a shared `src` folder, and on a Windows checkout without symlink support enabled, that symlink turns into a small stub file instead of a real link. That breaks asset loading inside the container. That is a separate, already known issue documented in this repo's README, and I am handling it properly in a following commit rather than papering over it here.

https://github.com/user-attachments/assets/577cf84f-8b79-476a-907e-f11ab068eed6

---

The problem was straightforward: there is no dev container in this repo, so nobody can use Codespaces or a one click "Reopen in Container" setup, and the Node version is only pinned in CI, not anywhere a local tool would read it.

For the base image, I considered writing a Dockerfile from scratch on top of a plain `node:22` image versus using the official Microsoft dev container image for Node. I chose the official image because it already ships a non-root user, which is the safer default and avoids a mistake I had to fix after the fact on a sibling repo's dev container. Since this app's test suite runs entirely under Vitest with jsdom, there was no need for any extra system packages like a browser or database, so the Dockerfile stays a single line.

The trickiest part was actually getting the app reachable. Everything looked correct on paper, config built, dependencies installed, but curling the forwarded port kept failing. I dug into it and found Vite only listens on localhost by default, which works fine when you run it directly on your own machine but does not work through a forwarded container port. Rather than changing that globally, which could have side effects I did not want to take responsibility for outside the dev container, I gated the change behind a `DEV_CONTAINER` variable that only the dev container sets, so nothing else is affected.

Key pieces:
- `.devcontainer/Dockerfile`: the Node 22 base image.
- `.devcontainer/devcontainer.json`: how the container is built and configured, the port it exposes, and the one editor extension it recommends.
- `vite.config.ts`: the one conditional line that makes the dev server reachable from outside the container.

I tested this by actually building the image, running it, confirming it runs as a non-root user on Node 22, running `npm ci` inside it, and starting the dev server, then curling the forwarded port from outside the container until it returned a real response instead of a connection failure.

---

---

## Summary by CodeRabbit

* **New Features**
  * Added dev container support for VS Code and GitHub Codespaces.
  * Configured automatic dependency installation, port 4000 forwarding, and recommended extensions.
  * Enabled development server access from containerized environments and Codespaces.

* **Documentation**
  * Added setup and usage instructions for the dev container.

* **Tests**
  * Added automated checks for container configuration and development server availability, including forwarded-host access and network binding.


tachyons:
Could you add video screencast as the demo ?

.github/workflows/devcontainer.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Build and smoke test dev container
+        uses: devcontainers/ci@v0.3
+        with:
+          push: never
+          runCmd: |
+            npm run lint:ci
+            npm run lint:v1:ci
+            npm run fmt:check
+            npm run test:src:ci
+            npm run test:v1:ci
```
Me-Priyank: All five commands already run on every PR (`lint.yml`, `vitest.yml`). Re-running them in a Node 22 container adds no container-specific signal and roughly triples CI time. Drop these and keep the build + `npm ci` + dev-server smoke test.

vite.config.ts
```diff
@@ -79,6 +79,7 @@ export default defineConfig(() => {
     },
     server: {
       port: 4000,
+      ...(process.env.DEV_CONTAINER ? { host: true } : {}),
```
Me-Priyank: Two fixes: `process.env.DEV_CONTAINER ?` is truthy for `="false"`, the Rails side uses `=== "true"` (`development.rb:78`); and `port: 4000` without `strictPort` means Vite silently falls back to 4001, which isn't forwarded and makes the smoke test fail misleadingly.

```ts
...(process.env.DEV_CONTAINER === "true" ? { host: true, strictPort: true } : {}),
```

Also, is a custom env var needed here at all? `npm run dev -- --host` needs no config change.

shivam-G05:
@tachyons have added the video file in the PR description

.github/workflows/devcontainer.yml
```diff
@@ -0,0 +1,43 @@
...
+
+            npm run dev > /tmp/dev.log 2>&1 &
+            for i in $(seq 1 30); do
+              code=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:4000/simulatorvue/v0/ || true)
```
Me-Priyank, line 44: This curls localhost from inside the same container, so it passes with or without `host: true`, the PR's only behavioural change is never exercised. Assert the bind address instead (`ss -ltn` showing `0.0.0.0`, not `127.0.0.1`).

.github/workflows/devcontainer.yml
```diff
@@ -0,0 +1,43 @@
+name: Dev Container
+
+on:
+  pull_request:
+    paths:
+      - ".devcontainer/*"
+
+jobs:
+  test-devcontainer:
+    runs-on: ubuntu-latest
```
Me-Priyank, line 20: Missing three conventions from `vitest.yml`: `permissions: {}`, a `concurrency` group, and `timeout-minutes`, the last matters most, since a `devcontainers/ci` hang burns the full 6-hour limit. `paths:` also doesn't cover `vite.config.ts`, `package*.json`, or this workflow.

vite.config.ts
shivam-G05: I checked this against how dev actually runs here. npm run dev maps to node scripts/dev.js, which spawns the underlying process as spawn('npx', ['vite'], { stdio: 'inherit', shell: true, env: process.env }). It never forwards process.argv to that child. The only CLI flag it special-cases is --desktop. So npm run dev -- --host doesn't actually reach Vite. I confirmed this live: Vite itself printed Network: use --host to expose and stayed bound to [::1]:4000 even with the flag passed.

Making --host work would mean changing scripts/dev.js to forward extra argv to the vite child process, which is a separate change and it also touches the desktop/Tauri dev path. Happy to do that as a later change if you'd rather standardize on the CLI flag, but for this PR I kept the DEV_CONTAINER env-var approach since it's the one that's actually wired up today. The two concrete fixes (=== "true" and strictPort: true) are in.

PR #1220, opened 2026-08-20
fix(testbench): resolve ReferenceError in v1 TestBench group navigation

### LINKED ISSUE #1219: 🐞 Bug: TestBench 'Next Group' navigation throws ReferenceError in v1 (undefined groupCount)

### Describe the bug

In the v1 (Vue) simulator, the TestBench panel's "Next Group" button (and, in turn, "Next Case" whenever it needs to roll over into the next group) throws an uncaught `ReferenceError` and never advances. Clicking it does nothing visible beyond an error in the console - group navigation in the TestBench is completely broken.

Root cause: `v1/src/simulator/src/testbench.ts`, `TestbenchData.groupNext()`:

```ts
groupNext() {
    const newCase = new TestbenchData(this.testData, this.currentGroup, 0);
    const _groupCount = newCase.testData.groups.length;   // declared as _groupCount
    let caseCount = 0;
    if (newCase.testData.groups[this.currentGroup].inputs[0]) {
      caseCount = newCase.testData.groups[this.currentGroup].inputs[0].values.length;
    }

    while (caseCount === 0 || this.currentGroup === newCase.currentGroup) {
      newCase.currentGroup++;
      if (newCase.currentGroup >= groupCount) return false;   // <-- references `groupCount`, not `_groupCount`
      caseCount = newCase.testData.groups[newCase.currentGroup].inputs[0].values.length;
    }
    ...
```

`groupCount` is never declared in this scope (or anywhere in the file) - only `_groupCount` is. Since `newCase.currentGroup` starts equal to `this.currentGroup`, the `while` loop's second condition (`this.currentGroup === newCase.currentGroup`) is always true on entry, so the loop body - and the reference to the undefined `groupCount` - executes on **every single call** to `groupNext()`, unconditionally.

### Steps to Reproduce

1. Run the v1 Vue simulator (`npm run dev -- v1`) with a circuit that has a TestBench with at least one test-case group.
2. Open the TestBench panel and click the "Next Group" button (`buttonListenerFunctions.nextGroupButton`, wired at `v1/src/components/Panels/TestBenchPanel/TestBenchPanel.vue:68`).
3. Open the browser console.
4. Observe `Uncaught ReferenceError: groupCount is not defined` thrown from `groupNext()`, and the TestBench view does not advance to the next group.
5. The same crash is reachable via the "Next Case" button (`nextCaseButton`) once the last case in a group is reached, since `caseNext()` calls `this.groupNext()` in that case (`testbench.ts:88`).

### Expected Behavior

Clicking "Next Group" (or reaching the end of a group via "Next Case") should advance `TestbenchData` to the next non-empty group without throwing, matching the working behavior in the v0 simulator (`src/simulator/src/testbench.ts`, which still declares this variable as `groupCount` and is unaffected).

### Root cause detail

Git blame traces this to commit `5328ea7` ("ci: introduce vue linter for v1 in ci"), which mechanically renamed several "unused" local variables to an underscore-prefixed name (a common convention to satisfy an ESLint `no-unused-vars`-style rule) without checking whether they were actually referenced elsewhere in the same function:

- `groupNext()`: `groupCount` -> `_groupCount`, but the loop body still reads `groupCount` (this is the crash reported here).
- `groupPrev()`: same rename, but `groupCount` isn't referenced anywhere in that function, so it's genuinely dead/unused there (harmless, but confirms the rename was applied mechanically rather than per-usage).
- `editTestButton()` (line ~653): `editDataString` -> `_editDataString`, also genuinely unused/dead code (a `JSON.stringify(...)` call whose result is discarded) - not a functional bug, just wasted work from the same commit.

The equivalent v0 file (`src/simulator/src/testbench.ts`) was not touched by that commit and still has the correctly-referenced `groupCount`, which is why this only affects v1.

### Impact

- The v1 TestBench's group navigation ("Next Group" button) is completely non-functional - every click throws.
- Any test-case flow that spans more than one group (a normal TestBench authoring pattern) is broken once the user reaches the end of a group and clicks "Next Case", since that also calls the broken `groupNext()`.
- Silent-ish failure: the error only surfaces in the browser console, not to the user, so it looks like the button simply does nothing.

### Affected components

- `v1/src/simulator/src/testbench.ts` - `TestbenchData.groupNext()` (the crash), plus the two harmless leftover unused variables noted above from the same commit (`groupPrev()`'s `_groupCount`, `editTestButton()`'s `_editDataString`).
- `v1/src/components/Panels/TestBenchPanel/TestBenchPanel.vue` - "Next Group" / "Next Case" buttons that trigger the crash.

### Acceptance criteria

- [ ] `groupNext()` no longer throws; clicking "Next Group" correctly advances to the next group with test cases (skipping empty groups), matching v0 behavior.
- [ ] `caseNext()` correctly rolls over into the next group when called on the last case of the current group.
- [ ] The genuinely-unused `_groupCount` in `groupPrev()` and `_editDataString` in `editTestButton()` are removed (dead code from the same commit) rather than left as unused underscore-prefixed variables.
- [ ] A regression test (unit test on `TestbenchData`) constructs test data with multiple groups (including at least one empty group to exercise the skip logic) and asserts `groupNext()`/`groupPrev()`/`caseNext()`/`casePrev()` advance correctly without throwing.

### Suggested solution direction

In `groupNext()`, change the undefined reference to use the already-declared count (renaming it back to `groupCount` since it is in fact used, or referencing `newCase.testData.groups.length` directly). Remove the two dead-code leftovers (`_groupCount` in `groupPrev()`, `_editDataString` in `editTestButton()`) rather than keeping unused variables around. This is a minimal, contained fix confined to `testbench.ts`.

Yes

### Vue simulator version

v1

Without Backend

Both

### Device Information

Not applicable - this is a plain JavaScript `ReferenceError`, reproducible in any browser.

### Additional Context

Confirmed present and unchanged on the current `upstream/main` HEAD. Confirmed absent from the v0 tree (`src/simulator/src/testbench.ts`), which still uses the correctly-referenced `groupCount`.

Yes

bhuvan-somisetty:
Fixes #1219

`TestbenchData.groupNext()` in `v1/src/simulator/src/testbench.ts` referenced an undefined variable `groupCount`. The variable was actually declared as `_groupCount` (an underscore-prefixed rename applied mechanically in commit `5328ea7` to silence an unused-variable lint warning), but that commit didn't update the one place inside the loop that reads it. Since `groupNext()`'s while-loop guard is always true on the first entry, this threw `ReferenceError: groupCount is not defined` on **every** call - the "Next Group" button in the v1 TestBench panel was completely non-functional, and clicking "Next Case" crashed the same way whenever it needed to roll into the next group (`caseNext()` calls `groupNext()` in that case).

Fix: renamed `_groupCount` back to `groupCount` in `groupNext()` (it is genuinely used). While tracing the same commit's blast radius, also removed two now-dead unused variables it left behind in the same file:
- `groupPrev()`'s `_groupCount` - never referenced anywhere in that function.
- `editTestButton()`'s `_editDataString` - a `JSON.stringify(...)` call whose result was discarded.

Both are pure dead-code removals with no behavior change; the only functional fix is the `groupNext()` rename.

Added `v1/src/simulator/spec/testbench.spec.ts` with regression tests covering `groupNext`/`groupPrev`/`caseNext`/`casePrev`, including empty-group skipping and first/last-group boundaries, so this class of bug (an undefined-variable typo surviving lint) is caught by `npm run test:v1` going forward.

N/A - this is a JS logic fix with no UI/markup changes. Before the fix, clicking "Next Group" in the TestBench panel threw an uncaught `ReferenceError` in the console and the view never advanced; after the fix it advances correctly.

---

The bug is a one-line typo: `git log -S _groupCount` traces it to commit `5328ea7` ("ci: introduce vue linter for v1 in ci"), which bulk-renamed apparently-unused locals to an underscore-prefixed name without checking each usage site. `groupNext()`'s loop body still reads `groupCount`, so the rename silently broke it - TypeScript didn't catch it, and it isn't exercised by any existing test since there was no spec file for `testbench.ts` before this PR.

I considered inlining `newCase.testData.groups.length` directly at the call site instead of restoring the named variable, but keeping the local `groupCount` matches the existing style (it mirrors `caseCount`, declared the same way just above) and keeps the diff minimal.

For the two other dead-code leftovers from the same commit (`groupPrev()`'s `_groupCount`, `editTestButton()`'s `_editDataString`), I verified they are genuinely unused anywhere in the file before removing them, rather than leaving unused variables in place - `groupPrev()` doesn't need an upper bound (it only checks `< 0` going backward), and `editTestButton()`'s stringified copy of the test data was never read before `openCreator("edit")` was called.

For tests, `TestbenchData` is a plain class with no DOM/circuit dependencies, but the rest of `testbench.ts` transitively imports `engine.js` -> `setup.js`, which side-effect-imports CodeMirror CSS that the vitest node runtime can't parse standalone. I mocked out the four unrelated imports (`engine`, `sequential`, `#/utils/confirm`, `#/simulator/src/utils`) that `TestbenchData` itself never touches, so the spec only exercises the class under test. I verified the tests actually catch the bug by running them against the pre-fix source (they fail with the exact `ReferenceError: groupCount is not defined` reported in the issue) before confirming they pass against the fix.

---

---

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for test navigation, including forward and backward movement, empty groups, group boundaries, and case transitions.
  * Added test scenarios with mocked dependencies and reusable test data.

* **Refactor**
  * Simplified internal navigation and test-editing logic without changing the public interface.


v1/src/simulator/spec/testbench.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  test("casePrev rolls back into the previous group's last case", () => {
+    const data = makeTestData([makeGroup("g0", 2), makeGroup("g1", 1)]);
+    const tb = new TestbenchData(data, 1, 0);
+
+    expect(tb.casePrev()).toBe(true);
+    expect(tb.currentGroup).toBe(0);
+    expect(tb.currentCase).toBe(1);
+  });
+});
```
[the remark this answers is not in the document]
bhuvan-somisetty: Added both, thanks — pushed in 5f1283f.

bhuvan-somisetty:
Addressed the CodeRabbit feedback — added the missing reverse-direction empty-group tests for `groupPrev()`/`casePrev()` so backward navigation across an empty group has the same coverage as the forward path. All 10 tests pass, lint and format checks are clean. Ready for another look.

v1/src/simulator/spec/testbench.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    expect(() => {
+      result = tb.groupNext();
+    }).not.toThrow();
+    expect(result).toBe(false);
+    expect(tb.currentGroup).toBe(0);
+  });
+});
+
+describe("TestbenchData.groupPrev", () => {
+  test("skips empty groups when moving back", () => {
```
Me-Priyank, line 94: Worth flagging: the four `groupNext` tests and `caseNext rolls over` are the ones that fail on the pre-fix source. These five backward-navigation tests pass either way, so they're characterisation coverage for previously untested code rather than regression coverage for this bug. Good to have, just not load-bearing for the fix.

v1/src/simulator/src/testbench.ts
```diff
@@ -69,7 +69,6 @@ export class TestbenchData {
 
   groupPrev() {
     const newCase = new TestbenchData(this.testData, this.currentGroup, 0);
-    const _groupCount = newCase.testData.groups.length;
```
Me-Priyank, line 72: Not blocking, and outside what this PR set out to do. `groupNext()` guards `.inputs[0]` before dereferencing, but this line and the two loop bodies (`groupNext()`, `groupPrev()`) don't. A group with `inputs: []` would `TypeError` here, and `removeTestButton` constructs exactly that shape. I couldn't find a click path that reaches the nav buttons in that state, since that handler also sets `showTestbenchUI = false`, so I'm happy to leave it. But `makeGroup` in the new spec always supplies one input, so nothing would catch it if that changed. Worth a follow-up issue.

Me-Priyank:
Verified all of it. The spec fails 5 of 10 on the pre-fix source with the exact `ReferenceError: groupCount is not defined`, passes 10/10 with the fix, the full v1 suite is green at 89 tests, and lint/format are clean. The blame to 5328ea7 checks out, and v0 does still declare `groupCount` correctly, so scoping this to v1 is right. Thanks for the "always true on entry" analysis of the loop guard, that's the detail that turns this from "sometimes broken" into "broken on every click", and it's correct.

LGTM

bhuvan-somisetty:
Thanks for the thorough verification, and for catching that empty-inputs edge case — opened #1231 to track it separately since it's not reachable today but worth guarding against.

tachyons (approved):

PR #1227, opened 2026-08-23
fix(combinational-analysis): drop redundant terms from minimized output

### LINKED ISSUE #1226: 🐞 Bug: Combinational Analysis returns a non-minimal expression with a redundant term
### Describe the bug

Combinational Analysis can return a minimized expression that contains a term already covered by the other terms, so the generated circuit has a redundant gate.

### Steps to reproduce

Open Combinational Analysis for 3 inputs (A, B, C) and set the output to 1 for the input rows 000, 001, 011 and 100.

Equivalently, calling the minimizer directly:

```js
new BooleanMinimize(3, [0, 1, 3, 4]).result
// ['00-', '0-1', '-00']   =>  A'B' + A'C + B'C'
```

### Expected behaviour

Two terms are enough, and the algorithm is supposed to pick the lowest cost cover:

- `A'C`  (`0-1`) covers 001 and 011
- `B'C'` (`-00`) covers 000 and 100

Those two already cover every minterm, so `A'B'` (`00-`) is redundant. The expected result is `A'C + B'C'`.

### Cause

In `src/simulator/src/quinMcCluskey.ts`, `get_essential_prime_implicants` builds the candidate covers:

```js
let append = true;

for (let j = tempSets.length - 1; j >= 0; --j) {
  if (isSubset(x, tempSets[j])) {
    tempSets.splice(j, 1);
  } else {
    append = false;
  }
}
```

The intent is to keep only the candidate covers that are not supersets of another. But the `else` fires whenever `x` is not a subset of `tempSets[j]`, which includes the case where the two are simply unrelated. An unrelated cover already sitting in `tempSets` is therefore enough to discard `x`, so valid candidates are lost and the search can no longer reach the minimal cover.

A candidate should be discarded only when an existing cover is contained in it.

### How widespread it is

Checking the minimizer against an independent exact search over prime implicants:

- every function of up to 4 variables is covered *correctly*, so no wrong outputs
- but many functions of 3 variables come back non minimal, for example minterms `[0,1,3,4]`, `[0,2,3,4]`, `[0,1,2,5]`, `[1,2,3,5]`, `[0,2,4,5]`
- 4 variable functions are affected too, for example `[0,1,2,9]` returns a cost 20 cover where 13 is achievable

So the output is always logically correct, it is just larger than it should be, which means extra gates in the generated circuit.

### Environment

Reproduced on `main` at `a7b9c89`.

rajanpanth:
Fixes #1226

Combinational Analysis could return a minimized expression containing a term already covered by the other terms, producing a redundant gate in the generated circuit.

For f(A, B, C) with minterms 0, 1, 3, 4:

```js
new BooleanMinimize(3, [0, 1, 3, 4]).result
// before: ['00-', '0-1', '-00']   A'B' + A'C + B'C'
// after:  ['0-1', '-00']          A'C + B'C'
```

`A'C` covers 001 and 011, `B'C'` covers 000 and 100, so those two already cover every minterm and `A'B'` adds nothing.

**Cause.** In `get_essential_prime_implicants`, the loop that keeps only the non-superset candidate covers reads:

```js
if (isSubset(x, tempSets[j])) {
  tempSets.splice(j, 1);
} else {
  append = false;
}
```

The `else` branch fires whenever `x` is not a subset of `tempSets[j]`, which includes the case where the two covers are merely unrelated. One unrelated cover already in `tempSets` was therefore enough to discard `x`, so valid candidates were dropped and the search could no longer reach the minimal cover.

**Fix.** Discard a candidate only when an existing cover is contained in it:

```js
if (isSubset(x, tempSets[j])) {
  tempSets.splice(j, 1);
} else if (isSubset(tempSets[j], x)) {
  append = false;
}
```

The output was never logically wrong, only larger than necessary, so this changes the size of the generated circuit rather than its behaviour.

None, the change is in the minimizer.

---

The problem is that the candidate-pruning step in the prime implicant cover search was too aggressive. `tempSets` is meant to hold the current minimal candidate covers, dropping any that are supersets of another. Subset and superset are not the only two possibilities though: two covers can be incomparable, and the original code lumped that third case in with "superset", throwing away candidates that were still needed.

The alternative I considered was leaving the pruning alone and instead scanning the final result for terms whose minterms are already covered, removing them at the end. That is a smaller change, but it only cleans up the specific symptom. The chosen fix repairs the search itself, so the algorithm evaluates all the candidate covers it was always meant to and the existing complexity comparison picks the cheapest, which also handles the cases where the cheapest cover is not simply the returned one minus a term.

The two key pieces are `isSubset`, which tests set containment, and the pruning loop that now distinguishes the three cases: `x` inside an existing cover means the existing one goes, an existing cover inside `x` means `x` is not worth keeping, and incomparable covers mean both are kept.

---

---

### Testing

Added `src/simulator/spec/quinMcCluskey.spec.ts`, which checks the minimizer against an independently computed reference rather than against its own current output:

- the specific `[0, 1, 3, 4]` regression
- every non-empty function of up to 4 variables covers exactly its minterms and nothing else
- every non-empty function of up to 3 variables reaches the same cost as an exact search over the prime implicants, enumerated independently in the test
- a don't care case

All four pass, in about 4 seconds. Reverting `quinMcCluskey.ts` while keeping the new spec fails 2 of the 4, so the tests exercise the change rather than restating current behaviour. The existing suite (`vitest run --project src`, 70 tests) still passes.

Note: `possibleSets = tempSets` on line 197 sits inside the `for (let s of possibleSets)` loop rather than after it. It happens to be harmless, since `for...of` holds the original iterator and `tempSets` keeps the same reference for the whole pass, so I left it alone to keep this diff focused. Happy to move it in a follow-up if you would like.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Boolean minimization to select accurate, minimal implicant covers.
  * Better preserves valid candidate solutions while removing redundant terms.
  * Improved handling of essential implicants, duplicate coverage entries, and don’t-care conditions.
  * Added safeguards for complex Boolean functions to keep minimization responsive.

* **Tests**
  * Added comprehensive validation for exact coverage and minimal solutions across Boolean functions with up to four variables.

### DIFF
--- a/src/simulator/spec/quinMcCluskey.spec.ts
+++ b/src/simulator/spec/quinMcCluskey.spec.ts
@@ -1,0 +1,155 @@
+import { describe, test, expect } from "vitest";
+import BooleanMinimize from "../src/quinMcCluskey";
+
+const bin = (n: number, width: number) => n.toString(2).padStart(width, "0");
+
+/** Whether a sum of products covers the given input assignment. */
+function evaluate(implicants: string[], input: string): boolean {
+  return implicants.some((term) =>
+    [...term].every((ch, i) => ch === "-" || ch === input[i]),
+  );
+}
+
+function minimize(numVars: number, minTerms: number[], dontCares: number[] = []): string[] {
+  return new (BooleanMinimize as any)(numVars, minTerms, dontCares).result;
+}
+
+/** The cost that solve() minimizes: one per term, per literal, and per inversion. */
+function complexity(terms: string[]): number {
+  let cost = terms.length;
+  for (const term of terms) {
+    for (const ch of term) {
+      if (ch !== "-") {
+        cost++;
+        if (ch === "0") cost++;
+      }
+    }
+  }
+  return cost;
+}
+
+/** Every prime implicant of the function, found by direct enumeration. */
+function primeImplicants(numVars: number, allowed: number): string[] {
+  const coverage = (term: string) => {
+    let mask = 0;
+    for (let i = 0; i < 1 << numVars; i++) {
+      if (evaluate([term], bin(i, numVars))) mask |= 1 << i;
+    }
+    return mask;
+  };
+
+  const terms: string[] = [];
+  for (let code = 0; code < 3 ** numVars; code++) {
+    let rest = code;
+    const chars: string[] = [];
+    for (let i = 0; i < numVars; i++) {
+      chars.push(["0", "1", "-"][rest % 3]);
+      rest = Math.floor(rest / 3);
+    }
+    const term = chars.join("");
+    if ((coverage(term) & ~allowed) === 0) terms.push(term);
+  }
+
+  return terms.filter((term) => {
+    const own = coverage(term);
+    return !terms.some((other) => other !== term && (coverage(other) & own) === own);
+  });
+}
+
+/**
+ * The lowest cost any sum of products can reach. A minimal cover never needs a
+ * non-prime implicant, since widening a term only removes literals.
+ */
+function optimalComplexity(numVars: number, minTerms: number[]): number {
+  let required = 0;
+  for (const m of minTerms) required |= 1 << m;
+
+  const primes = primeImplicants(numVars, required);
+  const covers = primes.map((term) => {
+    let mask = 0;
+    for (let i = 0; i < 1 << numVars; i++) {
+      if (evaluate([term], bin(i, numVars))) mask |= 1 << i;
+    }
+    return mask;
+  });
+
+  let best = Infinity;
+  const search = (index: number, covered: number, chosen: string[]) => {
+    const cost = complexity(chosen);
+    if (cost >= best) return;
+    if ((required & ~covered) === 0) {
+      best = cost;
+      return;
+    }
+    if (index >= primes.length) return;
+    search(index + 1, covered | covers[index], [...chosen, primes[index]]);
+    search(index + 1, covered, chosen);
+  };
+  search(0, 0, []);
+
+  return best;
+}
+
+/** Every minterm, and only the minterms, must be covered. */
+function coversExactly(result: string[], numVars: number, minTerms: number[]): boolean {
+  for (let i = 0; i < 1 << numVars; i++) {
+    if (evaluate(result, bin(i, numVars)) !== minTerms.includes(i)) return false;
+  }
+  return true;
+}
+
+/** All 2^(2^n) - 1 non-empty functions of n variables. */
+function* everyFunction(numVars: number): Generator<number[]> {
+  for (let mask = 1; mask < 1 << (1 << numVars); mask++) {
+    const minTerms: number[] = [];
+    for (let i = 0; i < 1 << numVars; i++) {
+      if (mask & (1 << i)) minTerms.push(i);
+    }
+    yield minTerms;
+  }
+}
+
+describe("BooleanMinimize", () => {
+  test("drops a term already covered by the others", () => {
+    // f = A'B' + A'C + B'C', where A'B' is redundant: A'C covers 001 and 011,
+    // B'C' covers 000 and 100, so together they cover every minterm.
+    expect(minimize(3, [0, 1, 3, 4])).toHaveLength(2);
+  });
+
+  test("covers exactly the minterms, for every function of up to four variables", () => {
+    const failures: string[] = [];
+
+    for (const numVars of [1, 2, 3, 4]) {
+      for (const minTerms of everyFunction(numVars)) {
+        const result = minimize(numVars, minTerms);
+        if (!coversExactly(result, numVars, minTerms)) {
+          failures.push(`[${minTerms}] => [${result}]`);
+        }
+      }
+    }
+
+    expect(failures.slice(0, 5)).toEqual([]);
+  });
+
+  test("finds a minimal cover, for every function of up to three variables", () => {
+    const suboptimal: string[] = [];
+
+    for (const numVars of [1, 2, 3]) {
+      for (const minTerms of everyFunction(numVars)) {
+        const result = minimize(numVars, minTerms);
+        const cost = complexity(result);
+        const best = optimalComplexity(numVars, minTerms);
+        if (cost > best) {
+          suboptimal.push(`[${minTerms}] => [${result}] costs ${cost}, best is ${best}`);
+        }
+      }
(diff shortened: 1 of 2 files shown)


src/simulator/src/quinMcCluskey.ts
```diff
@@ -181,10 +181,13 @@ BooleanMinimize.prototype.solve = function () {
...
           let append = true;
 
+          // Keep only the candidate covers that are not supersets of another.
+          // Two unrelated covers are both worth keeping, so `x` is dropped
+          // only when an existing cover is contained in it.
           for (let j = tempSets.length - 1; j >= 0; --j) {
             if (isSubset(x, tempSets[j])) {
               tempSets.splice(j, 1);
-            } else {
+            } else if (isSubset(tempSets[j], x)) {
```
Me-Priyank: The fix is correct, and I confirmed the spec fails 2 of 4 without it.

But the broken pruning was acting as an aggressive bound, and removing it exposes the full exponential cost of the cover enumeration. 

`combinationalAnalysis.js` allows up to 8 variables, and `GenerateCircuit` reads `temp.result` synchronously on the main thread at `combinationalAnalysis.js`. So this trades a slightly larger circuit for a multi-second freeze at 6 variables and an effective hang at 7-8.

The root cause is structural: despite the name, `get_essential_prime_implicants` never extracts essential PIs. It goes straight from the coverage table to enumerating every candidate cover. Adding the standard step first, take the PIs that uniquely cover a minterm, remove them and the minterms they cover, then run this search only on what's left, plus row/column dominance would keep almost all real inputs out of the exponential path entirely.

Could you pair the fix with that reduction?

src/simulator/spec/quinMcCluskey.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+describe("BooleanMinimize", () => {
+  test("drops a term already covered by the others", () => {
+    // f = A'B' + A'C + B'C', where A'B' is redundant: A'C covers 001 and 011,
+    // B'C' covers 000 and 100, so together they cover every minterm.
+    expect(minimize(3, [0, 1, 3, 4])).toHaveLength(2);
+  });
+
+  test("covers exactly the minterms, for every function of up to four variables", () => {
```
Me-Priyank, line 117: CodeRabbit asked for an explicit timeout here and 4dce5ae was oxfmt only, so this is still open. I measured this test at 1546 ms before the fix and 2036 ms after, against Vitest's 5000 ms default, passes locally, thin margin on a loaded runner. Please add `, 60_000`.

More importantly, the heaviest case in this file is 4 variables, and the performance cliff starts at 6. A bounded-time case at n=6-8 would have caught it, and would guard whatever reduction you add. Something like: minimize a dense 8-variable function and assert it completes under a second.

Me-Priyank:
The analysis is right and I reproduced every example in #1226 exactly, including the cost-20-vs-13 case for `[0,1,2,9]`. Validating against an independently computed reference instead of the implementation's own output is the right way to test this; note about `possibleSets = tempSets` is correct, `for...of` holds the original iterator, so the reassignment is harmless. Good catch to flag it.

Two things before this can merge, both in the inline comments: the performance regression at 6+ variables, and the still-open timeout request from CodeRabbit.

Also `v1/src/simulator/src/quinMcCluskey.ts` has the byte-identical block and isn't touched here, same as #1229. Please fix both copies, or say why v1 should be left.

And the AI-usage question in the template has neither option selected, again same as #1229. That one isn't optional.

rajanpanth:
You were right, and it was worse than the comment suggested. I measured before changing anything: the dense worst case (2 of every 3 minterms) did not finish in **60 seconds** at 6 variables, and crashed the Vitest worker at 553s when I let it run. So this was not a multi-second freeze, it was an effective hang, on the main thread.

Also confirmed your read of the structure: `get_essential_prime_implicants` never extracted an essential implicant. It went straight from the coverage table to enumerating candidate covers.

### What I did

Reduction first, as you asked:

- an implicant that uniquely covers a minterm is forced into every cover, so it is selected and its minterms dropped
- a row whose implicants are a superset of another row's is satisfied whenever that row is, so it is dropped
- rows are deduplicated, and the two steps repeat until nothing changes

That alone was not enough. On the dense case there are 60 prime implicants and no essentials at all, so nothing reduces and the full table still went to the enumeration: n=5 finished in 101ms, n=6 still did not finish in 60s.

So I also replaced the candidate enumeration with branch and bound: seed with a greedy cover, branch on the row with the fewest choices, prune any partial cover already at or above the best complete one. Cost is a sum over the chosen implicants, so that prune is sound.

Finding the true minimum here is set cover, so I capped the search rather than claim a bound I cannot deliver. Within the cap the answer is optimal; past it the greedy cover stands, which is still a correct cover, only possibly a term or two larger. I checked the cap is not doing damage on the hard case: at 200000, 50000 and 20000 steps the result was an identical 114 terms at n=8, so the extra search buys nothing there. Real truth tables reduce away long before the cap.

### Numbers

Dense worst case, `i % 3 !== 0`:

| n | before | after |
| --- | --- | --- |
| 6 | no finish in 60s | 112ms |
| 7 | no finish in 60s | ~250ms |
| 8 | no finish in 60s | 405ms |

### Tests

- added the bounded time case at 8 variables you asked for; it asserts the cover is exact and completes well inside the limit
- added explicit `60_000` timeouts to the two exhaustive tests
- the existing exhaustive checks still pass unchanged: every function of up to 4 variables is covered exactly, and every function of up to 3 variables still reaches the cost of an independently computed exact search

Reverting only `quinMcCluskey.ts` and keeping the specs makes the 8 variable test kill the worker, so it genuinely guards the regression.

`vitest run --project src --project v1` passes at 139 tests. `oxfmt --check` and `oxlint` are clean on both files, the two remaining warnings in that file (`BooleanMinimizeType`, `while (1)`) predate this PR.

One thing I did not do: column dominance. Essentials plus row dominance are unconditionally safe, whereas dropping an implicant because another covers a superset also has to compare cost to stay optimal, and the branch and bound already made the hard case fast. Happy to add it if you would rather have the table smaller before the search.

rajanpanth:
Friendly ping on this one. It has been approved and mergeable since 25 August with no failing checks, so I think it is just waiting for someone to press merge.

Happy to rebase if `main` has moved far enough that you would prefer a fresh run first.

PR #1229, opened 2026-08-23
fix(bit-converter): pass the decoded BCD value to setBaseValues

### LINKED ISSUE #1228: 🐞 Bug: BCD field in the bit converter throws ReferenceError and never updates the other fields
### Describe the bug

The BCD field in the Hex/Bin/Dec converter dialog does nothing. Typing into it throws `ReferenceError: x is not defined` and the other fields are never updated.

### Steps to reproduce

1. Open the bit converter dialog (Hex Bin Dec Converter).
2. Type a BCD value into the BCD field, for example `00100101`.
3. The decimal, binary, octal and hex fields stay unchanged. The browser console shows `ReferenceError: x is not defined`.

Every other field in the dialog works. Only the BCD one is affected.

### Cause

`setupBitConvertor` in `src/simulator/src/utils.ts`:

```js
$("#bcdInput").on("keyup", function () {
  var input = $("#bcdInput").val();
  var num = 0;
  while (input.length % 4 !== 0) {
    input = "0" + input;
  }
  if (input !== 0) {
    var i = 0;
    while (i < input.length / 4) {
      if (parseInt(input.slice(4 * i, 4 * (i + 1)), 2) < 10)
        num = num * 10 + parseInt(input.slice(4 * i, 4 * (i + 1)), 2);
      else return setBaseValues(NaN);
      i++;
    }
  }
  return setBaseValues(x);
});
```

The handler decodes the nibbles into `num`, then passes `x` to `setBaseValues`. There is no `x` in this scope. The other handlers each declare their own local `x`, so this looks like a copy of one of them where the variable name was not updated. The decoded `num` is discarded and the reference throws.

### Expected behaviour

`setBaseValues(num)`, so that entering `00100101` sets the decimal field to 25, hex to `0x19`, binary to `0b11001` and octal to `031`.

### Why the tests did not catch it

`spec/bitConvertor.spec.js` calls `setupBitConvertor()` and asserts it does not throw, which only registers the handlers. No test fires a `keyup` on the BCD field, so the body of the handler never runs.

### Environment

Reproduced on `main` at `a7b9c89`.

rajanpanth:
Fixes #1228

The BCD field in the Hex/Bin/Dec converter dialog did nothing. Typing into it threw `ReferenceError: x is not defined`, and the decimal, binary, octal and hex fields never updated.

In `setupBitConvertor`, the BCD keyup handler decodes the nibbles into `num` and then calls:

```js
return setBaseValues(x);
```

There is no `x` in that scope. The three other handlers in the same function each declare their own local `x`, so this reads as a copy of one of them where the variable name was not updated. The decoded value in `num` was discarded and the reference threw.

The fix passes `num`:

```diff
-    return setBaseValues(x);
+    return setBaseValues(num);
```

Entering `00100101` now sets decimal to 25, hex to `0x19`, binary to `0b11001` and octal to `031`.

No layout change. The BCD field simply updates the other fields now instead of throwing.

---

The handler already did the real work correctly. It left-pads the input so its length is a multiple of four, walks it four bits at a time, rejects any nibble that decodes to 10 or more since those are not valid BCD digits, and otherwise accumulates the digit into `num` with `num * 10 + nibble`. The only defect was the final line handing `setBaseValues` a variable that does not exist, so the one-word change is the whole fix.

I considered rewriting the handler to share a decode helper with the other three fields, but they each parse a different base with a one-line `parseInt`, whereas BCD needs the nibble walk, so there is not much genuinely shared. Keeping the change to the broken line also keeps it easy to review against the reported error.

One thing I deliberately left alone: `if (input !== 0)` compares a string against the number `0`, so it is always true and the guard does nothing. It is harmless, since an empty input skips the loop and yields 0 anyway, and changing it would alter behaviour beyond the reported bug. Happy to clean it up separately if you would like.

---

---

### Testing

The existing `spec/bitConvertor.spec.js` only asserted that `setupBitConvertor()` does not throw, which registers the handlers without ever running one, so the bug was invisible to it. Added two specs that actually fire `keyup` on the field:

- `00100101` updates all four other fields to 25 in their respective bases
- `1010` decodes to 10, which is not a valid BCD digit, so the fields keep their previous value

Both pass. Reverting `utils.ts` while keeping the specs reproduces `ReferenceError: x is not defined` in the first one, so it covers the change rather than restating current behaviour. The full suite (`vitest run --project src`) passes at 69 tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected BCD input handling so valid values accurately update decimal, hexadecimal, binary, octal, and normalized BCD displays.
  * Invalid BCD digits no longer overwrite previously displayed values.
  * Clearing the BCD input now preserves the existing converted values instead of resetting them.

* **Tests**
  * Added coverage for valid BCD conversions, cleared input behavior, and invalid BCD input handling.

### DIFF
--- a/src/simulator/spec/bitConvertor.spec.js
+++ b/src/simulator/spec/bitConvertor.spec.js
@@ -80,6 +80,26 @@ describe('data dir working', () => {
         expect(() => setupBitConvertor()).not.toThrow();
     });
 
+    test('bcd input updates the other fields', () => {
+        setupBitConvertor();
+        // 0010 0101 is the BCD encoding of the decimal number 25.
+        $('#bcdInput').val('00100101');
+        $('#bcdInput').trigger('keyup');
+        expect($('#decimalInput').val()).toBe('25');
+        expect($('#hexInput').val()).toBe('0x19');
+        expect($('#binaryInput').val()).toBe('0b11001');
+        expect($('#octalInput').val()).toBe('031');
+    });
+
+    test('bcd input rejects a nibble that is not a decimal digit', () => {
+        setupBitConvertor();
+        setBaseValues(7);
+        // 1010 is 10, which is not a valid BCD digit, so nothing should update.
+        $('#bcdInput').val('1010');
+        $('#bcdInput').trigger('keyup');
+        expect($('#decimalInput').val()).toBe('7');
+    });
+
     test('function setBaseValues working', () => {
         const randomBaseValue = Math.floor(Math.random() * 100);
         console.log('Testing for Base Value --> ', randomBaseValue);
--- a/src/simulator/spec/bitConvertor.spec.js
+++ b/src/simulator/spec/bitConvertor.spec.js
@@ -89,6 +89,9 @@ describe('data dir working', () => {
         expect($('#hexInput').val()).toBe('0x19');
         expect($('#binaryInput').val()).toBe('0b11001');
         expect($('#octalInput').val()).toBe('031');
+        // setBaseValues writes the field back through dec2bcd, which drops the
+        // leading nibble's zeros, so the typed 00100101 comes back as 100101.
+        expect($('#bcdInput').val()).toBe('100101');
     });
 
     test('bcd input rejects a nibble that is not a decimal digit', () => {
--- a/src/simulator/src/utils.ts
+++ b/src/simulator/src/utils.ts
@@ -288,7 +288,7 @@ export function setupBitConvertor() {
         i++;
       }
     }
-    return setBaseValues(x);
+    return setBaseValues(num);
   });
 
   $("#hexInput").on("keyup", function () {
--- a/v1/src/simulator/spec/bitConvertor.spec.js
+++ b/v1/src/simulator/spec/bitConvertor.spec.js
@@ -80,6 +80,29 @@ describe('data dir working', () => {
         expect(() => setupBitConvertor()).not.toThrow();
     });
 
+    test('bcd input updates the other fields', () => {
+        setupBitConvertor();
+        // 0010 0101 is the BCD encoding of the decimal number 25.
+        $('#bcdInput').val('00100101');
+        $('#bcdInput').trigger('keyup');
+        expect($('#decimalInput').val()).toBe('25');
+        expect($('#hexInput').val()).toBe('0x19');
+        expect($('#binaryInput').val()).toBe('0b11001');
+        expect($('#octalInput').val()).toBe('031');
+        // setBaseValues writes the field back through dec2bcd, which drops the
+        // leading nibble's zeros, so the typed 00100101 comes back as 100101.
+        expect($('#bcdInput').val()).toBe('100101');
+    });
+
+    test('bcd input rejects a nibble that is not a decimal digit', () => {
+        setupBitConvertor();
+        setBaseValues(7);
+        // 1010 is 10, which is not a valid BCD digit, so nothing should update.
+        $('#bcdInput').val('1010');
+        $('#bcdInput').trigger('keyup');
+        expect($('#decimalInput').val()).toBe('7');
+    });
+
     test('function setBaseValues working', () => {
         const randomBaseValue = Math.floor(Math.random() * 100);
         console.log('Testing for Base Value --> ', randomBaseValue);
--- a/v1/src/simulator/src/utils.ts
+++ b/v1/src/simulator/src/utils.ts
@@ -288,7 +288,7 @@ export function setupBitConvertor() {
         i++;
       }
     }
-    return setBaseValues(x);
+    return setBaseValues(num);
   });
 
   $("#hexInput").on("keyup", function () {


src/simulator/src/utils.ts
```diff
@@ -288,7 +288,7 @@ export function setupBitConvertor() {
         i++;
       }
     }
-    return setBaseValues(x);
+    return setBaseValues(num);
   });
```
Me-Priyank, line 292: Fix is correct, and I confirmed the new spec fails with `ReferenceError: x is not defined` when `utils.ts` is reverted, so it genuinely covers the change.

Two things before this merges:

`v1/src/simulator/src/utils` has the byte-identical bug and isn't touched here. Please apply the same fix (and ideally the same specs) to the v1 copy.

Also worth noting: with the handler now reaching `setBaseValues`, it writes back into `#bcdInput` via `convertors.dec2bcd`, which is `parseInt(x.toString(10), 16).toString(2)` with no zero-padding. Typing `00100101` leaves the field reading `100101`. The Vue component's `convertToBCD` uses `padStart(4, '0')` and doesn't have this problem. Not introduced by you, but the fix is what makes it reachable.

src/simulator/spec/bitConvertor.spec.js
```diff
@@ -80,6 +80,26 @@ describe('data dir working', () => {
...
+        // 0010 0101 is the BCD encoding of the decimal number 25.
+        $('#bcdInput').val('00100101');
+        $('#bcdInput').trigger('keyup');
+        expect($('#decimalInput').val()).toBe('25');
+        expect($('#hexInput').val()).toBe('0x19');
+        expect($('#binaryInput').val()).toBe('0b11001');
+        expect($('#octalInput').val()).toBe('031');
+    });
+
+    test('bcd input rejects a nibble that is not a decimal digit', () => {
```
Me-Priyank, line 97: This one passes both with and without the fix, since the invalid-nibble path returns at line 287 before reaching the line you changed. Good guard test to have, but only the `00100101` spec above actually covers the change.

could the first spec also assert `#bcdInput` after the keyup? It currently reads `100101` rather than the `00100101` that was typed, and pinning that down documents the padding gap.

Me-Priyank:
@rajanpanth thanks for working on this... root-cause analysis is right and the regression test earns its place. But the framing needs correcting.

`setupBitConvertor()` is never called anywhere in the app. `bitConverterDialog()` in `utils.ts` only sets `dialogBox.hex_bin_dec_converter_dialog = true`; the dialog that actually renders is `src/components/DialogBox/HexBinDec.vue` (mounted at `Extra.vue`), and its `bcdConverter()` already passes `num` correctly.

So the repro steps in #1228 are wrong. Step 1 opens the Vue dialog, not this jQuery handler; step 3's "browser console shows `ReferenceError: x is not defined`" cannot happen, because the code that throws is never bound to that field. The BCD field works in the running UI today, and the deploy preview on this PR shows it.

Please post the screen recording or screenshot of the actual bug that was asked for in the issue comment. If it can't be captured, that confirms this is a latent defect in unused legacy code rather than a user-facing break, and #1228 should be updated to say so. The fix is still worth taking, but the severity is different and we triage on that.

Also, the PR checklist is entirely unticked, including the AI-usage question where neither option is selected. That one isn't optional in this template.

Separately: this survived because nothing typechecks the repo. `vue-tsc` is in devDependencies but there's no `type-check` script or CI step, and `oxlint` doesn't flag undefined identifiers. `tsc` catches it instantly. Would you open a follow-up issue for that? It's the higher-value fix.

rajanpanth:
Thanks, you're right on all of it. I checked each point against the tree rather than taking it on trust:

- `setupBitConvertor()` is referenced only from `src/simulator/spec/bitConvertor.spec.js` and `v1/src/simulator/spec/bitConvertor.spec.js`. Nothing in application code binds that handler.
- `HexBinDec.vue`'s `bcdConverter()` ends in `return setBaseValues(num)`, so the rendered dialog is fine.

So the reproduction steps in #1228 cannot happen and there is no recording to post. I have corrected #1228 accordingly and suggested it be retitled to describe it as an undeclared identifier in unused legacy code rather than a broken dialog. The severity is yours to set, but it should not be triaged as user-facing.

Changes pushed:

- `v1/src/simulator/src/utils.ts` had the byte-identical line; fixed, with the same two specs added to the v1 project. Reverting only the v1 source fails the new v1 spec with `ReferenceError: x is not defined`.
- Both specs now assert `#bcdInput` after the keyup. It reads `100101`, not the `00100101` that was typed, because `setBaseValues` writes the field back through `dec2bcd`, which drops the leading nibble's zeros. Pinned with a comment so the padding gap is documented rather than implied.

You are also right that the invalid-nibble spec passes with and without the fix. It returns at the `NaN` branch before reaching the changed line, so it is a guard test, not coverage of this change. Only the `00100101` spec covers it. I have left both, but I am not claiming the guard test earns its place as coverage.

`vitest run --project src --project v1` passes at 138 tests.

Filed #1232 for the missing typecheck, which is the underlying reason this landed at all.

On the checklist: I have deliberately left the AI-usage section for the repository owner of this PR to complete rather than tick it myself, since those boxes are personal attestations about review and understanding.

rajanpanth:
On the CodeRabbit comment about empty BCD input: I'm not taking it, because it would make this copy diverge from the dialog that actually ships.

`HexBinDec.vue` has no empty-input guard either. `bcdConverter('')` leaves `num` at 0 and calls `setBaseValues(0)`, and `setBaseValues` there only returns early on `NaN`, so every field gets zeroed. That is exactly what `utils.ts` now does after this fix, so the two are consistent.

If zeroing on an empty field is the wrong behaviour, it is wrong in the shipped dialog, and fixing it there is a user-facing change that deserves its own PR and its own decision. Changing only the unused legacy copy would leave the two implementations disagreeing, which is how the `x` bug survived in the first place.

Same reasoning for the `if (input !== 0)` line, which compares a string against a number and is therefore always true. I flagged it in the description and left it deliberately: it is dead in both trees, and making it meaningful here alone would be the same divergence.

Happy to do the shipped-dialog version as a follow-up if you think the zeroing is wrong.

rajanpanth:
@Me-Priyank both of your points landed in `fbb0f4f`, pushed shortly after your review, so this may just need another look to clear the changes-requested state.

- **v1 copy**: `v1/src/simulator/src/utils.ts` had the byte-identical bug and now carries the same one-line fix, plus both specs mirrored into `v1/src/simulator/spec/bitConvertor.spec.js`.
- **Padding gap pinned**: the first spec now asserts `$('#bcdInput').val()` is `'100101'` after the keyup, with a comment explaining that `setBaseValues` writes back through `dec2bcd` and drops the leading nibble's zeros. It documents the gap rather than hiding it, as you suggested.

You were also right that the invalid-nibble spec passes with and without the fix, since that path returns before the changed line. I left it in as a guard test but it is not what covers the change; the `00100101` spec is. I verified that split rather than assuming it: reverting only `utils.ts` fails the `00100101` spec and leaves the invalid-nibble one green.

Both trees are passing: 10 tests across the two spec files.

I also left the CodeRabbit empty-input suggestion alone, for the reason in my earlier comment: `HexBinDec.vue`, the dialog that actually ships, has no such guard, so adding one here would make the two diverge.

rajanpanth:
Thanks, both points addressed, and one of them turned out to interact with #1247 in a way worth spelling out.

**The v1 copy.** You were right that it had the byte-identical bug. It has since been fixed on `main` directly, `v1/src/simulator/src/utils.ts` now reads `return setBaseValues(num)` along with the `String($("#bcdInput").val())` wrapper and `input !== "0"`. So after rebasing, the v1 side of this PR no longer carries the original one-line fix, only the empty-input guard below plus the specs. I kept your `String()` wrapper when resolving the conflict.

**The padding gap.** I did add the `#bcdInput` assertion you asked for, but not as a literal string, and the reason is that a literal would have made this PR and #1247 break each other. #1247 replaces `dec2bcd` with a per-digit `padStart(4, "0")` encoder, so the same input round-trips differently:

```
dec2bcd(25) on current main : 100101
dec2bcd(25) after #1247     : 00100101
```

Asserting `toBe('100101')` here would pass today and fail the moment #1247 lands, and vice versa if the order were reversed. The assertion is now width-independent:

```js
expect($('#bcdInput').val().replace(/^0+/, '')).toBe('100101');
```

I checked this rather than assuming: with #1247's `dec2bcd` pasted in locally, the spec still passes 6/6, and it passes 6/6 against current `main` too. So it documents the round-trip you wanted pinned down without either PR constraining the other.

**Empty input.** Reaching `setBaseValues` made one more edge case observable, which CodeRabbit also flagged. Clearing the field left `input` as `""`, the padding loop was a no-op because `"".length % 4 === 0`, `num` stayed `0`, and every field was rewritten to zero. Every sibling handler arrives at that state as `NaN` via `parseInt("")`, which `setBaseValues` ignores on line 247, so the BCD handler now does the same. Mutation check: with the guard removed and the new spec kept, it fails with `expected '0' to be '7'`.

**On your second point about the invalid-nibble spec:** agreed, it returns at the `else return setBaseValues(NaN)` branch before reaching the changed line, so it does not cover the fix. I left it in as a guard against that path regressing, but the `00100101` spec is the one that actually exercises the change.

Both trees are green: `Tests (src)` 6/6, `Tests (v1)` 6/6, `lint` and CodeQL passing, and `oxfmt --check` is clean.

PR #1238, opened 2026-08-28
fix(v1): stop TestbenchData.groupNext/groupPrev from crashing

### LINKED ISSUE #1231: v1 TestBench: groupNext/groupPrev can throw TypeError on groups with empty inputs
### Context

Follow-up from review discussion on #1220 (https://github.com/CircuitVerse/cv-frontend-vue/pull/1220#discussion_r3850370547).

`TestbenchData.groupNext()` guards `.inputs[0]` before dereferencing it, but this same dereference inside the `groupNext()`/`groupPrev()` loop bodies does not. A group with `inputs: []` would throw a `TypeError` there.

`removeTestButton` in `v1/src/simulator/src/testbench.ts` constructs exactly that shape (a group with an empty `inputs` array).

### Why this isn't urgent

No known click path currently reaches the nav buttons while a group is in that empty-inputs state — `removeTestButton`'s handler also sets `showTestbenchUI = false`, which hides the TestBench UI at the same time. So this is latent, not an active bug.

### Suggested fix

Add the same `.inputs[0]` existence guard used elsewhere in `groupNext()`/`groupPrev()`'s loop bodies, or otherwise skip/handle groups with no inputs when computing group counts for navigation.

Note: the regression spec added in #1220 (`v1/src/simulator/spec/testbench.spec.ts`) always supplies one input via `makeGroup`, so nothing currently catches this if it changes — a test case with an empty-`inputs` group should be added alongside the fix.

rajanpanth:
Fixes #1231

### What the issue asked for

`groupNext()` guards `inputs[0]` before its initial case-count read, but the same read inside both `groupNext()` and `groupPrev()`'s loop bodies does not. A group with no inputs (the shape `removeTestButton` constructs) throws a `TypeError` there.

### A second bug found while fixing it

Reading `groupNext()` to add the guard, its while loop references a bare `groupCount`:

```ts
const _groupCount = newCase.testData.groups.length;
...
if (newCase.currentGroup >= groupCount) return false;
```

The local is declared as `_groupCount`, underscore-prefixed, presumably to satisfy an unused-var lint rule after the loop body's own reference was left unrenamed. The class has no `groupCount` property either. I verified this in isolation before trusting it:

```
ReferenceError: groupCount is not defined
```

Since the while loop's first condition (`this.currentGroup === newCase.currentGroup`) is true on every call by construction, this throws whenever the loop needs even one iteration, which is effectively every call that has to skip forward past a group, not just the empty-inputs case the issue describes. `src/simulator/src/testbench.ts`'s implementation never had this typo (its local is plainly `groupCount`), which is how the v1 copy's mismatch shows up by comparison.

### Fix

- Renamed `_groupCount` back to `groupCount` in `groupNext()`, matching the working `src` copy.
- Added a `private static caseCountOf(group)` helper that returns `group.inputs[0] ? group.inputs[0].values.length : 0`, and routed every case-count read in `groupNext()`/`groupPrev()` through it, including the two loop-body reads the issue flagged and `groupPrev()`'s own initial read (also unguarded, not called out in the issue text but the same defect).

### Testing

No existing spec covers `TestbenchData` — `v1/src/simulator/spec/testbench.spec.ts` is new, per the issue's own note that nothing currently catches this. Eight cases: normal forward/backward navigation past a zero-case group, an empty-`inputs` group not throwing (start, intermediate, and every-remaining-group-empty positions) for both `groupNext` and `groupPrev`.

Reverting the source while keeping the tests fails 7 of 8, so they cover both bugs rather than restating current behaviour. `vitest run --project v1 --project src` passes at 142. `oxfmt --check` and `oxlint` are clean on both files.

Needed a small unrelated fix along the way: importing `testbench.ts` at all pulls in the verilog editor through `engine.ts`/`sequential.ts`, which imports codemirror's CSS, which Vitest can't transform. Mocked `codemirror` / `codemirror-editor-vue3` the same way `bitConvertor.spec.js` already does for the identical import chain.

### Note on scope

`inputs[0]` is dereferenced unguarded at several other points in this file (`isCaseValid`, `caseNext`, `casePrev`, `goToFirstValidGroup`'s later branch) that the issue does not mention and I have not verified are reachable with an empty-inputs group. Left those alone to stay inside what #1231 asked for; happy to open a follow-up if you'd like the whole class hardened at once.

---

The `groupCount` typo is the headline finding: renaming `_groupCount` back to `groupCount` in `groupNext()` makes it match the working `src` copy exactly, and I verified in isolation (a standalone reproduction of the class, no test framework) that the old code throws `ReferenceError: groupCount is not defined` the moment the loop needs to advance. The `.inputs[0]` guard is the fix the issue itself asked for; I factored it into one `caseCountOf()` helper used at every case-count read in both methods rather than repeating the ternary four times, since that is also what let me confirm I had not missed one of the loop-body reads.

I considered widening the fix to the other unguarded `inputs[0]` reads in the file (`isCaseValid`, `caseNext`, `casePrev`), but I have not verified those are reachable with an empty-inputs group the way `groupNext`/`groupPrev` are, so I left them out to stay inside what #1231 asked for and said so in the PR description.

---

- [ ] I have added proper PR title and linked to the issue
- [ ] I have performed a self-review of my code
- [ ] **I can explain the purpose of every function, class, and logic block I added**
- [ ] I understand why my changes work and have tested them thoroughly
- [ ] I have considered potential edge cases and how my code handles them
- [ ] My code follows the project's style guidelines and conventions

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved testbench navigation between groups with no test cases or inputs.
  - Prevented navigation errors after removing the final test case from a group.
  - Preserved next and previous group navigation for populated groups.

- **Tests**
  - Added coverage for navigating across empty, input-free, and populated testbench groups.
  - Verified navigation correctly reports when no eligible group is available.
  - Added checks for both forward and backward group navigation.

### DIFF
--- a/v1/src/simulator/spec/testbench.spec.ts
+++ b/v1/src/simulator/spec/testbench.spec.ts
@@ -1,0 +1,112 @@
+import { describe, expect, it, vi } from "vitest";
+import type { TestData } from "#/store/testBenchStore";
+
+// testbench.ts pulls in the verilog editor through engine.ts/sequential.ts,
+// which imports codemirror's CSS. Vitest can't transform a raw .css import,
+// so it is mocked the same way bitConvertor.spec.js does for the same chain.
+vi.mock("codemirror", async (importOriginal) => {
+  const actual = await importOriginal<Record<string, unknown>>();
+  return { ...actual, fromTextArea: vi.fn(() => ({ setValue: () => {} })) };
+});
+vi.mock("codemirror-editor-vue3", () => ({ defineSimpleMode: vi.fn() }));
+
+const { TestbenchData } = await import("../src/testbench");
+
+function makeGroup(values: string[]): TestData["groups"][number] {
+  return {
+    label: "g",
+    inputs: [{ label: "a", bitWidth: 1, values }],
+    outputs: [],
+    n: values.length,
+  };
+}
+
+function makeEmptyInputsGroup(): TestData["groups"][number] {
+  // The shape removeTestButton constructs: a group with no inputs at all.
+  return { label: "g", inputs: [], outputs: [], n: 0 };
+}
+
+function makeTestData(groups: TestData["groups"]): TestData {
+  return { type: "combinational", title: "t", groups };
+}
+
+describe("TestbenchData", () => {
+  describe("groupNext", () => {
+    it("advances past an intermediate group with no cases", () => {
+      const testData = makeTestData([makeGroup(["1"]), makeGroup([]), makeGroup(["1", "0"])]);
+      const tb = new TestbenchData(testData, 0, 0);
+
+      expect(tb.groupNext()).toBe(true);
+      expect(tb.currentGroup).toBe(2);
+    });
+
+    it("returns false when there is no next group with cases", () => {
+      const testData = makeTestData([makeGroup(["1"])]);
+      const tb = new TestbenchData(testData, 0, 0);
+
+      expect(tb.groupNext()).toBe(false);
+    });
+
+    it("does not throw on a group with no inputs, and skips past it", () => {
+      const testData = makeTestData([
+        makeGroup(["1"]),
+        makeEmptyInputsGroup(),
+        makeGroup(["1", "0"]),
+      ]);
+      const tb = new TestbenchData(testData, 0, 0);
+
+      expect(() => tb.groupNext()).not.toThrow();
+      expect(tb.currentGroup).toBe(2);
+    });
+
+    it("returns false rather than throwing when every later group has no inputs", () => {
+      const testData = makeTestData([makeGroup(["1"]), makeEmptyInputsGroup()]);
+      const tb = new TestbenchData(testData, 0, 0);
+
+      expect(() => tb.groupNext()).not.toThrow();
+      expect(tb.groupNext()).toBe(false);
+    });
+  });
+
+  describe("groupPrev", () => {
+    it("moves back past an intermediate group with no cases", () => {
+      const testData = makeTestData([makeGroup(["1"]), makeGroup([]), makeGroup(["1", "0"])]);
+      const tb = new TestbenchData(testData, 2, 0);
+
+      expect(tb.groupPrev()).toBe(true);
+      expect(tb.currentGroup).toBe(0);
+    });
+
+    it("does not throw when the starting group has no inputs", () => {
+      const testData = makeTestData([makeGroup(["1"]), makeEmptyInputsGroup()]);
+      const tb = new TestbenchData(testData, 1, 0);
+
+      let moved: boolean | undefined;
+      expect(() => {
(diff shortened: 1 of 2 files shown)


JituRewar (commented):
@rajanpanth The same assertion issue appears in both the `groupNext` and `groupPrev` "all remaining groups are empty" tests.

`expect(() => tb.groupNext()).toBe(false)` / `expect(() => tb.groupPrev()).toBe(false)` asserts that the callback itself equals `false`, rather than checking the callback's return value.

Could these be changed to:

```ts
expect(tb.groupNext()).toBe(false);
expect(tb.groupPrev()).toBe(false);
```

The `not.toThrow()` assertion is already sufficient for the separate no-throw check.

rajanpanth:
Thanks @JituRewar — done in `2588cef`, both tests now read exactly as you suggested.

One clarification on the diagnosis, since it changes what was actually wrong: the branch didn't contain `expect(() => tb.groupNext()).toBe(false)`. It had the no-throw check and the value check as two separate lines:

```ts
expect(() => tb.groupNext()).not.toThrow();
expect(tb.groupNext()).toBe(false);
```

So the assertion wasn't vacuous. But your conclusion still holds for a different reason: those are two separate calls, and `groupNext()` mutates state, so the `toBe(false)` was asserting the return of the *second* invocation rather than the first. Collapsing to the single line you proposed removes the double call, which is the stronger test.

The no-throw coverage is unaffected, since a throw fails the remaining assertion anyway. I confirmed that rather than assuming it: reverting only `testbench.ts` to its pre-fix state still fails 7 of the 8 tests, including both of the ones I just changed, with the original `TypeError: Cannot read properties of undefined`. All 8 pass with the fix in place.

rajanpanth:
Rebased onto current `main` and the conflict is resolved, so this is mergeable again.

Worth flagging what the conflict was, since `main` has moved in the same area. A commit on `main` now guards the first `caseCount` read in `groupNext`:

```ts
let caseCount = 0;
if (newCase.testData.groups[this.currentGroup].inputs[0]) {
  caseCount = newCase.testData.groups[this.currentGroup].inputs[0].values.length;
}
```

but the read inside the loop just below it is still unguarded:

```ts
caseCount = newCase.testData.groups[newCase.currentGroup].inputs[0].values.length;
```

so a group with no inputs encountered while scanning forward still throws, and `groupPrev` is untouched. I kept this branch's `caseCountOf` helper, which covers all four read sites in both methods and is equivalent to the new guard for the first one.

Verified against current `main` rather than assuming: reverting only `testbench.ts` to `origin/main` and keeping this branch's spec still fails 5 of the 8 tests, including "does not throw on an intermediate group with no inputs" and both "returns false rather than throwing" cases. With the helper, all 8 pass.

PR #1260, opened 2026-09-05
Fix(vuesim): add conditional check for propagationDelay when its set for 0

### LINKED ISSUE #1259: 🐞 Bug: [vuesim] Propagation delay in large circuits

### Describe the bug

In case of large circuits, like the [32-bit cpu circuit](https://circuitverse.org/users/92698/projects/string-32000), there is a significant delay which can be observed.

### Steps to Reproduce

1. enable vuesim in the flipper
2. load the [32-bit cpu](https://circuitverse.org/users/92698/projects/string-32000)
3. any characters entered in the keyboard buffer doesnt get drained 

### Expected Behavior

Comparing this with the legacy sim, the buffer drains perfectly after every character typed. This is a delay issue which is caused in this circuit

### Screenshots

_No response_

Yes

### Vue simulator version

v0

Without Backend

Both

### Device Information

```markdown

```

### Additional Context

_No response_

Yes

ShinichiShi:
Fixes #1259 

 

This PR addresses the issue where in the keyboard buffer was not being drained for the [32000 circuit](https://circuitverse.org/users/92698/projects/string-32000). This issue was caused mainly due to the propagation delay of the circuit not being handled properly in the `load.js` file when compared with legacy sim.

---

Root cause: 

In the `load.js` file, the `data.propagationDelay` is the delay value the circuit's author saved for that element. 
and this line : `obj.propagationDelay = data.propagationDelay || obj.propagationDelay` makes 0 as false in JS, 
Therefore an element deliberately saved with `propagationDelay`: 0 was silently promoted to the class default (10) on every load. 

Legacy's loader file already guarded against this by adding the condition but this fix was not migrated to vuesim too.

In this particular circuit case: 
The event queue schedules every element at time = currentTime + delay. Circuits that rely on a cluster of zero-delay elements settling within the same simulated instant (e.g. STRING32000, a CPU that executes one instruction per clock edge) get that cluster's relative ordering against the rest of the circuit shifted by +10 on load. For a feedback loop, that's enough to permanently prevent part of it from ever re-entering the queue, no exceptions, just a branch that silently never resolves again. Concretely, this is why STRING 32000.cv's keyboard buffer filled but never drained in the Vue simulator while working fine in legacy.

Therefore, restored the same conditional guard legacy already has in both versions of vuesims. 

---

---

Credits: 
Huge thanks to @aryan7071 for helping out entirely in identifying this particular fix :raised_hands: 

## Summary by CodeRabbit

- **Bug Fixes**
  - Saved propagation delays of zero are now preserved correctly when loading simulator data.
  - Prevents zero-delay configurations from being unintentionally replaced by an existing delay value.

### DIFF
--- a/src/simulator/src/data/load.js
+++ b/src/simulator/src/data/load.js
@@ -56,7 +56,11 @@ function loadModule(data, scope) {
         data.labelDirection || oppositeDirection[fixDirection[obj.direction]]
 
     // Sets delay
-    obj.propagationDelay = data.propagationDelay || obj.propagationDelay
+    if (data.propagationDelay === 0) {
+        obj.propagationDelay = 0;
+    } else {
+        obj.propagationDelay = data.propagationDelay || obj.propagationDelay;
+    }
     obj.fixDirection()
 
     // Restore other values
--- a/v1/src/simulator/src/data/load.js
+++ b/v1/src/simulator/src/data/load.js
@@ -56,7 +56,11 @@ function loadModule(data, scope) {
         data.labelDirection || oppositeDirection[fixDirection[obj.direction]]
 
     // Sets delay
-    obj.propagationDelay = data.propagationDelay || obj.propagationDelay
+    if (data.propagationDelay === 0) {
+        obj.propagationDelay = 0
+    } else {
+        obj.propagationDelay = data.propagationDelay || obj.propagationDelay
+    }
     obj.fixDirection()
 
     // Restore other values


ShinichiShi:
This fix was made in the legacy sim before in this [PR:#3928](https://github.com/CircuitVerse/CircuitVerse/pull/3928)

Nihal4777 (approved):
Great work guys!! 🚀

